Issue #105, opened 2015-06-17
Compile option to disable all Lua support


leto:
I would like to ask if there is a single ./configure flag to disable all Lua, since currently I am using something like:

```
./configure -j2 --with-debug --with-http_postgres_module --with-pcre  \
                --with-http_realip_module                             \
                --without-lua51                                       \
                --without-lua_cjson                                   \
                --without-lua_redis_parser                            \
                --without-lua_rds_parser                              \
                --without-lua_resty_dns                               \
                --without-lua_resty_memcached                         \
                --without-lua_resty_redis                             \
                --without-lua_resty_mysql                             \
                --without-lua_resty_upload                            \
                --without-lua_resty_string                            \
                --without-lua_resty_websocket                         \
                --without-lua_resty_lock                              \
                --without-lua_resty_lrucache                          \
                --without-lua_resty_core                              \
                --without-http_autoindex_module                       \
                --with-http_auth_request_module               \
                --without-http_ssi_module                             \
                --without-lua_resty_upstream_healthcheck
```

which is less than awesome and needs to be kept in sync with every command-line option related to Lua. If there is not something like this, than please point me in the correct direction and I will submit a PR

agentzh:
@leto You need to add these too: `--without-http_lua_module --without-http_lua_upstream_module`.

No, there's no such all-in-one option to disable all Lua. Feel free to submit a PR :) Not sure about the option name though. My ears are open :)

leto:
@agentzh how about ```without-any-lua``` ?

agentzh:
@leto Looks good to me :)

leto:
@agentzh great! I will prepare a PR in my fork at https://github.com/letolabs/ngx_openresty . Is there a document that has guidelines for developer contributions that I am not seeing?

agentzh:
@leto Not yet. I meant to work on such a document in the future :) Well, just go ahead and create the PR. I'll review it anyway.

PR #106, opened 2015-06-17
Add a --without-any-lua configure option

leto:
This is related to #105

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -154,7 +154,25 @@ for my $opt (@ARGV) {
     if ($opt =~ /^--prefix=(.*)/) {
         $prefix = $1;
 
-    } elsif ($opt eq '--without-lua51') {
+    } elsif ($opt eq '--without-any-lua') {
+        $resty_opts{no_lua_cjson}                      = 1;
+        $resty_opts{no_lua_redis_parser}               = 1;
+        $resty_opts{no_lua_resty_memcached}            = 1;
+        $resty_opts{no_lua_resty_redis}                = 1;
+        $resty_opts{no_lua_resty_dns}                  = 1;
+        $resty_opts{no_lua_resty_mysql}                = 1;
+        $resty_opts{no_lua_resty_upload}               = 1;
+        $resty_opts{no_lua_resty_string}               = 1;
+        $resty_opts{no_lua_resty_websocket}            = 1;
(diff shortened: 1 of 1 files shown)


agentzh:
Hmm, seems like we also need to disable LuaJIT. Before that, we need to have a `--without-luajit` option. Thoughts?

leto:
@agentzh I can add that too! Would you prefer it be in this PR or a different one?

agentzh:
@leto Either is fine :) Thanks!

agentzh:
@leto I have a mixed feeling about this. Is an OpenResty without any Lua still OpenResty?

bungle:
This is for advanced users, and having a config option(s) to disable things is well, just flexible. I'm against having `--without-any-lua`. You can disable them one-by-one, :-). If you disable/remove Lua altogether, maybe then you are prohibited to call it OpenResty anymore?

leto:
My original reason for this was that I used OpenResty for a non-Lua project, since it makes it so easy to  use various add-ons. I was mostly using rds_json and postgres modules to implement a JSON API directly in Nginx/OpenResty, without any dynamic language. It was extremely performant.

You can think of this as an "easy on-ramp" for people to use and get interesting in OpenResty without Lua. I ended up having 20 ``--without`` statements in my provision scripts to compile OpenResty, which is why I suggested ```--without-any-lua```.

bungle:
@leto,

I know, `without-any-lua` sounds a bit harsh, maybe:

```
--without-lua-libraries
```

Is a bit better. That could include all `lua-resty-*`and `lua-cjson` and `lua-rds-parser`.

And possibly (these are rather static though):

```
--without-lua-modules
```

agentzh:
@leto @bungle  I think a better approach is to automatically exclude all the Lua libraries and the ngx_http_lua_upstream module when the user specifies the `--without-http_lua_module` option. This looks more reasonable to me. What do you guys think?

bungle:
@agentzh, that's even better. Doesn't get any easier than than.

agentzh:
Patches welcome! Volunteers welcome! :)

bungle:
@leto, would you like to modify your patch so that `--without-http_lua_module` disables all the Lua-related?

leto:
@bungle yep, gonna work on it today!

PR #177, opened 2016-05-21
feature: add an option --add-resty-module

doujiang24:
just like nginx standard `--add-module`


agentzh:
I think it's better to let the upcoming `opm` command-line utility to handle local installations of such Lua modules for OpenResty, as in

```bash
opm install /path/to/lua-resty-foo/
```

doujiang24:
understand

PR #197, opened 2016-09-14
Use sbin-path when creating symlink, fixes #196

### LINKED ISSUE #196: Not using sbin-path configuration when creating symlink
Should the symlink be created using whatever path was set with "--sbin-path" instead of assuming "sbin/nginx"?
https://github.com/openresty/openresty/blob/fd667d63695c640791791e86997778e9b081786d/util/configure#L347

davidgaleano:


agentzh:
@davidgaleano Mind to add a test case to the existing test suite to cover this change? Thanks!

davidgaleano:
@agentzh  Apologies, I was not aware of the existing tests, I have updated them to work with my proposed changes.

The fixes for the existing test cases are more extensive that I thought because I also changed the configure script to always respect the DESTDIR when creating the soft link, which was something required for our own build to work because we do set custom values for DESTDIR, sbin-path, and prefix.

t/sanity.t
```diff
@@ -315,7 +315,7 @@ install: all
 	cp -r $OPENRESTY_BUILD_DIR/pod $(DESTDIR)/usr/local/openresty/
 	cd $OPENRESTY_BUILD_DIR/nginx-1.11.2 && $(MAKE) install DESTDIR=$(DESTDIR)
 	mkdir -p $(DESTDIR)/usr/local/openresty/site/lualib $(DESTDIR)/usr/local/openresty/site/pod $(DESTDIR)/usr/local/openresty/site/manifest
-	ln -sf /usr/local/openresty/nginx/sbin/nginx $(DESTDIR)/usr/local/openresty/bin/openresty
+	ln -sf $(DESTDIR)/usr/local/openresty/nginx/sbin/nginx $(DESTDIR)/usr/local/openresty/bin/openresty
```
agentzh: This looks wrong to me and it may break OpenResty own official Linux packages here:

http://openresty.org/en/linux-packages.html

util/configure
```diff
@@ -354,7 +359,10 @@ push @make_install_cmds,
     . " \$(DESTDIR)$prefix/site/manifest";
 
 if ($platform ne 'msys') {
-    push @make_install_cmds, "ln -sf $ngx_prefix/sbin/nginx \$(DESTDIR)$prefix/bin/openresty";
+    if (!$ngx_sbin) {
+        $ngx_sbin = "$ngx_prefix/sbin/nginx";
+    }
+    push @make_install_cmds, "ln -sf \$(DESTDIR)$ngx_sbin \$(DESTDIR)$prefix/bin/openresty";
```
agentzh: Will you explain this change? Apparently this create wrong symlinks in the RPM packaging process, for example.

t/sanity.t
davidgaleano: Ah! yes, our internal packages are not like the official linux packages. Our packages are installed on non-standard locations, an all the executables are located on the same directory. I will change it to not break things, and somehow make our build system do what we need...

util/configure
davidgaleano: Yes, this is just an artifact of how we build and install our packages, I will remove the $(DESTDIR) prefix on the source.

agentzh:
@davidgaleano Looking good to me. Thanks!

@doujiang24 @membphis @moonbingbing @thibaultcha Will you please double check? Thanks!

thibaultcha:
Looking good to me, plus tests passing on my side 

@davidgaleano Gj :+1:

agentzh:
@thibaultcha Thanks for your review.

@davidgaleano Just merged. Thanks!

PR #214, opened 2016-12-17
feature: applied the safe_resolver_ipv6_option patch 

thibaultcha:
As discussed in openresty/resty-cli#20, this patch to the nginx core provides a failsafe when the `ipv6=off` option is given in the `resolver` directive while nginx is not built with IPv6 support. It also provides a somewhat nicer message when `ipv6=on` without support for IPv6.

I tested this on my side with the `util/mirror-tarballs` utility and then installing the given tarball, everything seemed ok. Let me know if anything else is missing!


agentzh:
@thibaultcha Looking good to me :)

@yangshuxin @doujiang24 @membphis @moonbingbing Mind to have a quick look at this?

doujiang24:
LGTM:)

agentzh:
@thibaultcha Merged. Thanks!

PR #215, opened 2016-12-17
save few bytes on response

chipitsine:


agentzh:
@chipitsine We want to avoid patching the nginx core wherever possible. If you think it's useful, please propose it to the nginx-devel mailing list instead. Generally, openresty only includes nginx core patches that are necessary.

chipitsine:
I could not convince Maxim Dounine to accept that patch

If you can convince nginx guy to accept it, it will be nice

agentzh:
@chipitsine No, I can't either :)

chipitsine:
as that patch is somewhat usefull, why not to keep it within openresty source tree ?

chipitsine:
I use that patch myself about 3 years. like 200k users (varoious browsers, etc). I beleive it would be useful for Cloudflare as well

agentzh:
@chipitsine Some notes:

1. I no longer work for Cloudflare.
2. For CDN vendors and other large scale users, they also need to accommodate somewhat incompatible HTTP clients, not just well-behaved web browsers. So we need to be conservative for any protocol-level changes.
3. As I've already pointed out, we want to minimize the number of patches for nginx core in OpenResty.

PR #235, opened 2017-03-03
feature: applied the resolve hostsfile patch to the nginx core

thibaultcha:
This patch allows the Nginx resolver to parse the system hostsfile
(/etc/hosts) when specifying a `hostsfile` option to the resolver
directive.

Would love to get some feedback on it.

Patch originally discussed at openresty/resty-cli#10.

This patch was also proposed to the nginx-devel mailing list:
http://mailman.nginx.org/pipermail/nginx-devel/2017-February/009544.html
It is slightly revised compared to the one proposed there.

### DIFF
--- a/patches/nginx-1.11.2-resolve-hostsfile.patch
+++ b/patches/nginx-1.11.2-resolve-hostsfile.patch
@@ -1,0 +1,373 @@
+# HG changeset patch
+# User Thibault Charbonnier <thibaultcha@me.com>
+# Date 1488252201 28800
+#      Mon Feb 27 19:23:21 2017 -0800
+# Branch resolve-hostsfile
+# Node ID e348d382e1ed78fba55261c1962625415ea21fd6
+# Parent  640f035293959b2d4b0ba5939d954bc517f57f77
+Resolver: parse hosts file entries
+
+The resolver directive can now take an optional 'hostsfile=<path>' option, such
+as:
+
+    resolver 8.8.4.4 hostsfile=/etc/hosts;
+
+Hosts parsed from the hosts file are considered valid forever. The behavior
+tries to be conservative, and only parses the hosts file when the option is
+provided, to enforce backwards compatibility.
+
+diff -r 640f03529395 -r e348d382e1ed src/core/ngx_resolver.c
+--- a/src/core/ngx_resolver.c	Fri Jan 27 19:06:35 2017 +0300
++++ b/src/core/ngx_resolver.c	Mon Feb 27 19:23:21 2017 -0800
+@@ -9,11 +9,12 @@
+ #include <ngx_core.h>
+ #include <ngx_event.h>
+ 
+-
+-#define NGX_RESOLVER_UDP_SIZE   4096
+-
+-#define NGX_RESOLVER_TCP_RSIZE  (2 + 65535)
+-#define NGX_RESOLVER_TCP_WSIZE  8192
++#define NGX_RESOLVER_HOSTSFILE_SIZE  4096
++
++#define NGX_RESOLVER_UDP_SIZE        4096
++
++#define NGX_RESOLVER_TCP_RSIZE       (2 + 65535)
++#define NGX_RESOLVER_TCP_WSIZE       8192
+ 
+ 
+ typedef struct {
+@@ -120,6 +121,8 @@
+     ngx_resolver_node_t *rn);
+ static void ngx_resolver_srv_names_handler(ngx_resolver_ctx_t *ctx);
+ static ngx_int_t ngx_resolver_cmp_srvs(const void *one, const void *two);
++static ngx_int_t ngx_resolver_parse_hostsfile(ngx_conf_t *cf,
++    ngx_resolver_t *r);
+ 
+ #if (NGX_HAVE_INET6)
+ static void ngx_resolver_rbtree_insert_addr6_value(ngx_rbtree_node_t *temp,
+@@ -128,7 +131,6 @@
+     struct in6_addr *addr, uint32_t hash);
+ #endif
+ 
+-
+ ngx_resolver_t *
+ ngx_resolver_create(ngx_conf_t *cf, ngx_str_t *names, ngx_uint_t n)
+ {
+@@ -243,6 +245,24 @@
+         }
+ #endif
+ 
++        if (ngx_strncmp(names[i].data, "hostsfile=", 10) == 0) {
++            r->hostsfile.len = names[i].len - 10;
++
++            if (r->hostsfile.len == 0) {
++                ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
++                                   "invalid parameter: %V", &names[i]);
++                return NULL;
++            }
++
++            r->hostsfile.data = ngx_resolver_dup(r, names[i].data + 10,
++                                                 r->hostsfile.len + 1);
++            if (r->hostsfile.data == NULL) {
++                return NULL;
++            }
++
++            continue;
++        }
++
+         ngx_memzero(&u, sizeof(ngx_url_t));
+ 
+         u.url = names[i];
+@@ -273,6 +293,13 @@
+         }
+     }
+ 
++    if (r->hostsfile.len > 0
++        && ngx_resolver_parse_hostsfile(cf, r)
++        != NGX_OK)
++    {
++        return NULL;
++    }
++
+     return r;
+ }
+ 
+@@ -4660,3 +4687,265 @@
+ 
+     return p1 - p2;
+ }
++
++
++static ngx_int_t
++ngx_resolver_parse_hostsfile(ngx_conf_t *cf, ngx_resolver_t *r)
++{
++    off_t                    file_size;
++    u_char                   ch;
++    u_char                  *start;
++    size_t                   len;
++    ssize_t                  n, size;
++    ngx_int_t                rc;
++    ngx_buf_t                b;
++    ngx_fd_t                 fd;
++    ngx_str_t                filename, s;
++    ngx_file_t               file;
++    in_addr_t                addr;
++    ngx_resolver_node_t     *rn;
++    enum {
++        scan_line = 0,
++        scan_skipline,
++        scan_addr,
++        scan_hosts,
++        scan_name
++    } state;
++
++    b.start = NULL;
++    s.data = NULL;
++    s.len = 0;
++    rn = NULL;
++    rc = NGX_OK;
++
++    filename = r->hostsfile;
++
++    fd = ngx_open_file(filename.data, NGX_FILE_RDONLY, NGX_FILE_OPEN, 0);
++    if (fd == NGX_INVALID_FILE) {
++        ngx_log_error(NGX_LOG_ALERT, cf->log, ngx_errno,
++                      ngx_open_file_n " \"%s\" failed", filename.data);
++        return NGX_ERROR;
++    }
++
++    ngx_memzero(&file, sizeof(ngx_file_t));
++
++    if (ngx_fd_info(fd, &file.info) == NGX_FILE_ERROR) {
++        ngx_log_error(NGX_LOG_ALERT, cf->log, ngx_errno,
++                      ngx_fd_info_n " \"%s\" failed", filename.data);
++        goto fail;
++    }
++
++    file.fd = fd;
++    file.log = cf->log;
++    file.name.len = filename.len;
++    file.name.data = filename.data;
++    file.offset = 0;
++
++    b.start = ngx_resolver_alloc(r, NGX_RESOLVER_HOSTSFILE_SIZE);
++    if (b.start == NULL) {
++        goto fail;
++    }
++
++    b.pos = b.start;
++    b.last = b.start;
++    b.end = b.last + NGX_RESOLVER_HOSTSFILE_SIZE;
++    b.temporary = 1;
++
++    start = b.pos;
++    state = scan_line;
++    file_size = ngx_file_size(&file.info);
++
++    for ( ;; ) {
++
++        if (b.pos >= b.last) {
++            len = b.pos - start;
++
++            if (len) {
++                ngx_memmove(b.start, start, len);
++            }
++
++            size = (ssize_t) (file_size - file.offset);
++
++            if (size > b.end - (b.start + len)) {
++                size = b.end - (b.start + len);
++
++            } else if (size == 0) {
++                goto done;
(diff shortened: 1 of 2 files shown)


agentzh:
@thibaultcha I'm afraid it's usually not desired to make the entries never expired :) Otherwise everytime /etc/hosts is edited, we need to reload nginx. Maybe it's better to expose a pure C API callable by FFI that can insert any records into the shm dns cache used by the nginx internal resolver? We can do whatever we want on the Lua land, including periodic updates in timers and parsing any external stuff (not just `/etc/hosts`).

thibaultcha:
> Otherwise everytime /etc/hosts is edited, we need to reload nginx.

I believed this would actually be preferred. Here is the reasoning:
- There is no TTL definition in the hosts file, hence, I believe most people expect that editing this file instantly takes effect.
- However, if users expect that editing the hostsfile instantly updates their records, then we would be fooling them, since that would mean we'd need to constantly parse the file, which we don't (and won't?) do.
- Hence, we'd rather promise them that the records will be parsed and considered valid for the lifetime of their workers. If users are editing their hostsfile and expecting the changes to take effect, they should reload their workers as well.

Ultimately, I could see a manually-specified TTL value for the hosts file records, and us reloading the file via AIO (if possible?). I however do believe this would be overkill, and as just described, slightly confusing to our users.

> Maybe it's better to expose a pure C API callable by FFI that can insert any records into the shm dns cache used by the nginx internal resolver? We can do whatever we want on the Lua land, including periodic updates in timers and parsing any external stuff (not just /etc/hosts).

I think this would be useful as well and is a good idea. However, wouldn't it be a different work/scope than this one? I can see both this patch and this Lua API live together :)

agentzh:
@thibaultcha Yeah, your patch has its own right. But I don't think it's reasonable to require specifying an external nameserver when enabling the hosts file. Maybe it's better to be an isolate configuration directive? The hosts file are parsed at compile-time anyway and has no logical relationship with the request-time resolver anyway (though it has implementation relationships, which do not count :)).

agentzh:
@thibaultcha Something like this maybe?

```nginx
hosts_file /etc/hosts;
```

thibaultcha:
> hosts_file /etc/hosts;

Yes, I remember this is how you envisioned this, and I considered it for this implementation :) 

However, I am concerned about a few things if we do so:

1. The size and scope of the patch. If I am correct, this would be the first significant patch applied to nginx core, and it would add a new directive (unheard of in our core patches) instead of an option for a directive. I believe such a patch would be more sensitive to future changes from Nginx core? It would touch several files, including `ngx_http_core_module`
2. I am not sure how well the nginx resolver behaves when no name servers are configured? This definitely requires some testing and possible some more patching...
3. We would need to re-implement the `ipv6={on,off}` parsing logic in this new `hosts_file` directive. This smells of code duplication, and sounds like a bad sign for the future. Our parsing of this option already differs from nginx core (remember my previous patch? 😅)

I will do some research on 2. to eventually switch to `hosts_file` if you do not change your mind, but let me know if I am making sense or not :)

Thanks!

agentzh:
@thibaultcha Yeah, the change would be much larger. I just think grouping hosts file into the resolver directive has a bad design smell and counter-intuitive.

My personal feeling is that maybe we should do the resolver cache C API for the Lua land, which can do everything this patch can do (though requiring a little bit of Lua), and only require a presumably very small change in the nginx core. And this also has no impact on nginx core's configuration UI and thus no forward compatibility concerns. Just my 2 cents :)

thibaultcha:
Ok, that is settled then, I'll look into implementing this as `hosts_file` :)

I am definitely on your side for the resolver cache Lua API as well, and comfortable implementing it. I just think that since we can already resolve hostnames thanks to lua-resty-dns, the hosts file parsing currently would offer more value to our users. Just my 2 cents as well but I might be wrong!

thibaultcha:
Closing this in the meanwhile.

thibaultcha:
Introducing `hosts_file` really creates a lot of code duplication and almost doubles the patch size. We now need to handle such `hosts_file /etc/hosts ipv6=off;` cases as well, so we duplicate the parsing code, and we also need to introduce this new directives in all three of the http, stream, and eventually mail modules, which might be too much.

However, I am a big fan of not having to specify a name server for such hosts file resolution, and have updated the resolver to work without name servers with minimal changes. How about specifying a resolver like so:
```
resolver hostsfile=/etc/hosts ipv6=off;
```

This would keep the patch at about the same size as this one and also answers your concern. Tell me your thoughts :)

thibaultcha:
Let me know what you think of my latest update @agentzh :)

thibaultcha:
@agentzh I don't feel good about double-pinging (sorry), but I 👂 word of a 1.11.2.3 release and was wondering if the merge window was closed already, and if not, wanted to know if there was a way this could sneak in given the above update. Many thanks!

agentzh:
@thibaultcha Sorry for the delay on my side.

@dndx @doujiang24 What do you guys think of this patch?

agentzh:
@thibaultcha The merge window for OpenResty 1.11.2.3 is already closed. Sorry about that. Want to kick out a new release ASAP so that we can start working on compatibilities with the latest nginx core.

dndx:
This patch looks fine to me (aside from a minor issue above). It will be even better if we could also write some simple tests for it. (maybe inside https://github.com/openresty/openresty/tree/master/t)

agentzh:
Oh, yeah, we need test cases for it.

@thibaultcha Mind to write some quick tests with `Test::Nginx`? ;) We could put it in this repo, as @dndx suggested.

thibaultcha:
> Mind to write some quick tests with Test::Nginx?

Yes, yes of course, for sure :) I wasn't sure where to include those originally.

> aside from a minor issue above

@dndx Hi! Do you mind clarifying which issue you are referring to here? Is it the missing ability to define a resolver without a nameserver, or an issue in the current patch submitted with this PR?

patches/nginx-1.11.2-resolve-hostsfile.patch
```diff
@@ -0,0 +1,373 @@
...
++            r->hostsfile.len = names[i].len - 10;
++
++            if (r->hostsfile.len == 0) {
++                ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
++                                   "invalid parameter: %V", &names[i]);
++                return NULL;
++            }
++
++            r->hostsfile.data = ngx_resolver_dup(r, names[i].data + 10,
++                                                 r->hostsfile.len + 1);
```
dndx: Hmm, looks like `ngx_conf_read_token()` actually [adds the `'\0'` for directive parameters](http://lxr.nginx.org/source/src/core/ngx_conf_file.c#0760). That is definitely not something I knew before! Closing this review.

dndx:
@thibaultcha When running your patch with the no_pool patch using Valgrind, I got the following warning:

```
==7725== 11 bytes in 1 blocks are definitely lost in loss record 2 of 12
==7725==    at 0x4A06C50: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==7725==    by 0x42F4B9: ngx_alloc (ngx_alloc.c:22)
==7725==    by 0x421194: ngx_resolver_alloc (ngx_resolver.c:4130)
==7725==    by 0x42132B: ngx_resolver_dup (ngx_resolver.c:4176)
==7725==    by 0x422DD8: ngx_resolver_create (ngx_resolver.c:257)
==7725==    by 0x4394A4: ngx_http_core_resolver (ngx_http_core_module.c:5105)
==7725==    by 0x41FC07: ngx_conf_handler (ngx_conf_file.c:462)
==7725==    by 0x41FC07: ngx_conf_parse (ngx_conf_file.c:318)
==7725==    by 0x437072: ngx_http_block (ngx_http.c:237)
==7725==    by 0x41FC07: ngx_conf_handler (ngx_conf_file.c:462)
==7725==    by 0x41FC07: ngx_conf_parse (ngx_conf_file.c:318)
==7725==    by 0x41D477: ngx_init_cycle (ngx_cycle.c:275)
==7725==    by 0x40C4C4: main (nginx.c:284)
```

Why don't we change

```
r->hostsfile.data = ngx_resolver_dup(r, names[i].data + 10,
                                     r->hostsfile.len + 1);
```

to

```
r->hostsfile.data = names[i].data + 10;
```

instead? I can confirm that fixes the warning. AFAIK we don't actually need to `dup` the string here.

thibaultcha:
@dndx Nice, thank you! I'll look into it. And into making the `resolver` work without nameserver (I have that somewhere...)

dndx:
@thibaultcha Found another issue.

When building `nginx` with `--with-debug` and this patch, and `nginx.conf` contains:

```
resolver hostsfile=/etc/hosts;
```

Valgrind will output messages similar to this:

```
[datong@localhost nginx-build]$ valgrind ./sbin/nginx -t
==14154== Memcheck, a memory error detector
==14154== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==14154== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==14154== Command: ./sbin/nginx -t
==14154== 
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
==14154== Warning: invalid file descriptor -1 in syscall write()
nginx: the configuration file /home/datong/nginx-build/conf/nginx.conf syntax is ok
nginx: configuration file /home/datong/nginx-build/conf/nginx.conf test is successful
```

`strace` confirms that `nginx` is attempting to write `malloc()` debug logs to fd `-1`.

```
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 0000000002225280:9\n", 64) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F1B0:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F270:9\n", 64) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F290:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F350:21\n", 65) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F370:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F430:10\n", 65) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F450:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F510:23\n", 65) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 0000000002225280:3\n", 64) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F530:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F5F0:9\n", 64) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F610:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F6D0:21\n", 65) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F6F0:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F7B0:10\n", 65) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F7D0:184\n", 66) = -1 EBADF (Bad file descriptor)
write(-1, "2017/04/05 16:47:20 [debug] 14157#0: malloc: 000000000223F890:23\n", 65) = -1 EBADF (Bad file descriptor)
close(5)
```

Commenting out `resolver hostsfile=/etc/hosts;` will get rid of this warning.

I did some digging and it appears to have something to do with your call to `ngx_resolver_dup()` inside this patch. I believe that `r->log` (which corresponds to `cf->cycle->new_log`) have not been initialized inside `ngx_resolver_create()` and as a result, `ngx_resolver_dup()` which passes it as the log facility to `ngx_alloc()` will attempt to write into an uninitialized file descriptor.

I also did some digging using GDB and can confirm that `ngx_log_open_default()` which opens the file is called after your code returns.

I think we should just use `ngx_alloc()` or `ngx_calloc()` directly [as shown in above code](http://lxr.nginx.org/source/src/core/ngx_resolver.c#0149).

thibaultcha:
Sweet, thanks. I'll reproduce and fix on my side as well!

thibaultcha:
Unfortunately, I force pushed a new version of the patch before reopening this PR, and GitHub won't let me do so anymore. I have opened #247 instead. Sorry!

PR #247, opened 2017-04-06
feature: applied the resolve hosts file patch to the nginx core

thibaultcha:
**Reopened from #235.** Unfortunately, I force pushed a new version of the 
patch before reopening #235, and GitHub won't let me do so anymore.
cc. @agentzh @dndx

This patch allows the Nginx resolver to parse the system hosts file
(/etc/hosts) when specifying an hosts file option to the resolver
directive.

TODO: 
- [ ] tests in `t/`

Patch originally discussed at openresty/resty-cli#10.

This patch was also proposed to the nginx-devel mailing list:
http://mailman.nginx.org/pipermail/nginx-devel/2017-February/009544.html
It is slightly revised compared to the one proposed there.


thibaultcha:
@agentzh @dndx Updated the patch with tests.

- tests in `t/patches`
- updated the patch to handle a resolver without nameservers
- slightly improved the parsing logic to handle empty lines

Q: now, `prove t/` will run both the `sanity.t` and the `patches/*.t` tests. However, the later requires a compiled OpenResty, and not just the bundle like the former. How would like to to handle that case, or structure those tests differently maybe?

Thanks!

patches/nginx-1.11.2-resolve-hosts-file.patch
```diff
@@ -0,0 +1,413 @@
...
++                {
++
++#if (NGX_HAVE_INET6)
++                    if (!r->ipv6
++                        || ngx_inet6_addr(s.data, s.len,
++                                          rn->u6.addr6.s6_addr) != NGX_OK)
++                    {
++                        ngx_resolver_free_node(r, rn);
++                        state = scan_skipline;
++                        break;
```
thibaultcha: Here, we are missing such a `break` when the parsed value is an ipv6 and when Nginx is built *without* ipv6 support. Will update.

t/patches/01-resolve-hosts-files.t
```diff
@@ -0,0 +1,165 @@
...
+
+    location = /t {
+        echo "Hello";
+    }
+--- request
+GET /t
+--- response_body
+Hello
+--- no_error_log
+host "localhost-v6" will resolve to "::1"
```
thibaultcha: This test will fail when Nginx is built without ipv6 for the reason highlighted above for now.

thibaultcha:
@agentzh @dndx Alright, I have updated the patch/tests to pass with and without ipv6 support. Do you mind giving it another round of review sometime? Thanks!

thibaultcha:
I also tested this in Valgrind mode. Do we have any more thoughts on it? Should we setup a Travis CI instance to run those new tests that need a compiled OpenResty? (+ the old ones requiring a bundled OpenResty tarball).

thibaultcha:
@agentzh Is this still in the picture? I'll start the work on `ngx.resolver` as well once I can focus on it. I still think this patch would be nice to have...

agentzh:
@thibaultcha Yeah, I agree it is a good thing to have.

Your patch generally looks good to me. Will have a closer look and test it out myself once I have some more time. Been very busy with other higher priority things.

bernhardreiter:
Just curious, what is missing here to get this merged?

PR #249, opened 2017-04-07
feature: Intercepting NGINX error logs with custom hooks.

membphis:
I didn't reuse `log->write`, we should write the log to file if the intercepting fails, e.g: buffer is full.
but the return value of `log->write` is `void`.


patches/nginx-1.11.2-hook_error_log.patch
```diff
@@ -0,0 +1,86 @@
+diff --git a/src/core/ngx_log.c b/src/core/ngx_log.c
+index 8e9408d..093351c 100644
+--- a/src/core/ngx_log.c
++++ b/src/core/ngx_log.c
+@@ -164,6 +164,13 @@ ngx_log_error_core(ngx_uint_t level, ngx_log_t *log, ngx_err_t err,
+             break;
+         }
+ 
++        if (log->intercept_log_handler && log->enter_error_log == 0)
++        {
```
agentzh: Style: `{` should be on the previous line.

patches/nginx-1.11.2-hook_error_log.patch
```diff
@@ -0,0 +1,86 @@
...
++
+     ngx_log_writer_pt    writer;
+     void                *wdata;
+ 
+@@ -70,6 +74,8 @@ struct ngx_log_s {
+     char                *action;
+ 
+     ngx_log_t           *next;
++
++    unsigned             enter_error_log:1;
```
agentzh: The field name `entered_logger` is better.

Use the following line:

```
    unsigned                entered_logger;  /* :1 */
```

There is no point in using `:1` here for the current layout (with padding considered).

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,86 @@
...
+     (void) ngx_atomic_fetch_add(ngx_stat_reading, 1);
+diff --git a/src/http/ngx_http_request.h b/src/http/ngx_http_request.h
+index 499c1ef..6c7ca9f 100644
+--- a/src/http/ngx_http_request.h
++++ b/src/http/ngx_http_request.h
+@@ -383,6 +383,7 @@ struct ngx_http_request_s {
+     ngx_http_upstream_t              *upstream;
+     ngx_array_t                      *upstream_states;
+                                          /* of ngx_http_upstream_state_t */
++    ngx_array_t                      *intercept_logs;    /* of ngx_str_t */
```
agentzh: Why do we need this field?

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,86 @@
...
+diff --git a/src/http/ngx_http_core_module.h b/src/http/ngx_http_core_module.h
+index 773c215..3be0158 100644
+--- a/src/http/ngx_http_core_module.h
++++ b/src/http/ngx_http_core_module.h
+@@ -175,6 +175,8 @@ typedef struct {
+     ngx_uint_t                 try_files;       /* unsigned  try_files:1 */
+ 
+     ngx_http_phase_t           phases[NGX_HTTP_LOG_PHASE + 1];
++
++    ngx_log_intercept_pt       intercept_log_handler;
```
agentzh: Why do we need this field? I think the log intercept handler should apply to all the subsystems, not just `http`.

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,86 @@
...
++
+     ngx_log_writer_pt    writer;
+     void                *wdata;
+ 
+@@ -70,6 +74,8 @@ struct ngx_log_s {
+     char                *action;
+ 
+     ngx_log_t           *next;
++
++    unsigned             enter_error_log:1;
```
agentzh: There is no point in using `:1` here since there's no other fields can fill up the padding it saves. So it will just waste CPU time while saving *no space*.

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,86 @@
+diff --git a/src/core/ngx_log.c b/src/core/ngx_log.c
+index 8e9408d..093351c 100644
+--- a/src/core/ngx_log.c
++++ b/src/core/ngx_log.c
+@@ -164,6 +164,13 @@ ngx_log_error_core(ngx_uint_t level, ngx_log_t *log, ngx_err_t err,
+             break;
+         }
+ 
++        if (log->intercept_log_handler && log->enter_error_log == 0)
++        {
```
agentzh: Again, `{` should be on the previous line. Be careful.

patches/nginx-1.9.15-hook_error_log.patch
membphis: It now only supports "http" scenes.

We intercepted the error logs from `ngx_log_error_core` and save it to `ngx_http_request_s->intercept_logs`, save the intercept log to the current request structure.

It's clearer, and its easy to control the max number of error log per request. It's not safe if we save it as worker level(eg: one request generates a large number of duplicate error logs).

patches/nginx-1.11.2-hook_error_log.patch
membphis: fixed

patches/nginx-1.9.15-hook_error_log.patch
doujiang24: @agentzh I also think buffer the error messages to request level is better for now.
Since we need to drop the error messages when a request is done but still have error messages belong to it left in the worker level buffer, which is a bit difficult here.
@membphis I personally think limit the max number of error log per request is not so necessary here.

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
++        log_intercept = ngx_cycle->log_intercept_handler;
++    }
++
+     while (log) {
+ 
+         if (log->log_level < level && !debug_connection) {
+             break;
+         }
+ 
++        if (log_intercept && log->entered_logger == 0) {
```
agentzh: Because we use `entered_logger` as a boolean value, we should test `!log->entered_logger` instead.

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
++            log->entered_logger = 1;
++            log_intercept(log, level, errstr, p - errstr);
```
agentzh: Seems like this is in the wrong place. We would get duplicate error log messages when the current log contains multiple loggers. Note that this is inside a `while` loop.

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
+     ngx_uint_t   wrote_stderr, debug_connection;
+     u_char       errstr[NGX_MAX_ERROR_STR];
+ 
++    ngx_log_intercept_pt    log_intercept;
++
+     last = errstr + NGX_MAX_ERROR_STR;
+ 
+     p = ngx_cpymem(errstr, ngx_cached_err_log_time.data,
+@@ -158,12 +160,23 @@ ngx_log_error_core(ngx_uint_t level, ngx_log_t *log, ngx_err_t err,
+     wrote_stderr = 0;
```
agentzh: I think the intercept log handler should be called before the line

```C
    ngx_linefeed(p);
```

Also, it's better to record the log level integer directly at the same time. We should let the log data consumer to decide what level to filter, so that we can implement highly dynamic level filtering without reloading nginx.

patches/nginx-1.9.15-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
+     debug_connection = (log->log_level & NGX_LOG_DEBUG_CONNECTION) != 0;
+ 
++    log_intercept = NULL;
```
agentzh: I think we can initialize it to `NULL` in the variable declaration above instead.

patches/nginx-1.11.2-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
++        log_intercept = ngx_cycle->log_intercept_handler;
++    }
++
+     while (log) {
+ 
+         if (log->log_level < level && !debug_connection) {
+             break;
+         }
+ 
++        if (log_intercept && log->entered_logger == 0) {
```
agentzh: `entered_logger` is used as a boolean value, so we should test `!log->entered_logger` instead.

patches/nginx-1.11.2-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
++typedef ngx_int_t (*ngx_log_intercept_pt) (ngx_log_t *log, ngx_uint_t level,
++    u_char *buf, size_t len);
+ 
+ struct ngx_shm_zone_s {
+     void                     *data;
+@@ -75,6 +77,7 @@ struct ngx_cycle_s {
+     ngx_str_t                 prefix;
+     ngx_str_t                 lock_file;
+     ngx_str_t                 hostname;
++    ngx_log_intercept_pt      log_intercept_handler;
```
agentzh: I think `intercept_error_log_handler` is a better name.

agentzh: Also, we should introduce a special macro like `HAVE_INTERCEPT_ERROR_LOG_PATCH`, similar to our existing `HAVE_ALLOW_REQUEST_BODY_UPDATING_PATCH` macro.

patches/nginx-1.11.2-hook_error_log.patch
```diff
@@ -0,0 +1,71 @@
...
+diff --git a/src/core/ngx_log.h b/src/core/ngx_log.h
+index 385fee4..57e3795 100644
+--- a/src/core/ngx_log.h
++++ b/src/core/ngx_log.h
+@@ -70,6 +70,8 @@ struct ngx_log_s {
+     char                *action;
+ 
+     ngx_log_t           *next;
++
++    unsigned             entered_logger;    /* :1 */
```
agentzh: I think we can just put this field in `ngx_cycle_t` as well, right after the `intercept_error_log_handler` field. There's no point in allocating another 8 bytes for every `ngx_log_t` struct in every request on a 64 bit system.

membphis:
@agentzh fixed all point, and removed nginx `1.9.15` patch.  

commit: https://github.com/openresty/openresty/pull/249/commits/30854e6ce338856c561ea4c07a5608431e5c69db

patches/nginx-1.11.2-hook_error_log.patch
```diff
@@ -0,0 +1,63 @@
...
++    ngx_log_intercept_pt    log_intercept = NULL;
++
+     last = errstr + NGX_MAX_ERROR_STR;
+ 
+     p = ngx_cpymem(errstr, ngx_cached_err_log_time.data,
+@@ -153,6 +155,20 @@ ngx_log_error_core(ngx_uint_t level, ngx_log_t *log, ngx_err_t err,
+         p = last - NGX_LINEFEED_SIZE;
+     }
+ 
++#ifdef HAVE_INTERCEPT_ERROR_LOG_PATCH
```
agentzh: We don't need this guard in this patch itself.

agentzh:
@membphis Just merged with the following extra patch:

```diff
diff --git a/patches/nginx-1.11.2-hook_error_log.patch b/patches/nginx-1.11.2-intercept_error_log.patch
similarity index 100%
rename from patches/nginx-1.11.2-hook_error_log.patch
rename to patches/nginx-1.11.2-intercept_error_log.patch
diff --git a/util/mirror-tarballs b/util/mirror-tarballs
index 4063980..29350d3 100755
--- a/util/mirror-tarballs
+++ b/util/mirror-tarballs
@@ -55,9 +55,19 @@ else
     fi
 fi

-echo "$info_txt applying the privileged-agent-process patch"
-patch -p1 < $root/patches/nginx-$main_ver-privileged_agent_process.patch || exit 1
-echo
+answer=`$root/util/ver-ge "$main_ver" 1.11.2`
+if [ "$answer" = "Y" ]; then
+    echo "$info_txt applying the privileged-agent-process patch"
+    patch -p1 < $root/patches/nginx-$main_ver-privileged_agent_process.patch || exit 1
+    echo
+fi
+
+answer=`$root/util/ver-ge "$main_ver" 1.11.2`
+if [ "$answer" = "Y" ]; then
+    echo "$info_txt applying the intercept-error-log patch"
+    patch -p1 < $root/patches/nginx-$main_ver-intercept_error_log.patch || exit 1
+    echo
+fi

 echo "$info_txt applying the upstream-pipelining patch for nginx"
 patch -p1 < $root/patches/nginx-$main_ver-upstream_pipelining.patch || exit 1
```

PR #250, opened 2017-04-15
feature: supported privilege process.

membphis:
The privilege process has the same permission as Nginx `master` process, and it closed the listening socket for safety.

We can do something special in this privilege process.

### DIFF
--- a/patches/nginx-1.11.2-privilege_process.patch
+++ b/patches/nginx-1.11.2-privilege_process.patch
@@ -40,20 +40,8 @@ index c51b7ff..3261f90 100644
  
      ngx_msec_t                timer_resolution;
  
-diff --git a/src/os/unix/ngx_process.c b/src/os/unix/ngx_process.c
-index 24a63fb..4579aba 100644
---- a/src/os/unix/ngx_process.c
-+++ b/src/os/unix/ngx_process.c
-@@ -396,6 +396,7 @@ ngx_signal_handler(int signo)
- 
-         break;
- 
-+    case NGX_PROCESS_PRIVILEGED:
-     case NGX_PROCESS_WORKER:
-     case NGX_PROCESS_HELPER:
-         switch (signo) {
 diff --git a/src/os/unix/ngx_process_cycle.c b/src/os/unix/ngx_process_cycle.c
-index 3ead164..6151e8a 100644
+index 3ead164..4e06e97 100644
 --- a/src/os/unix/ngx_process_cycle.c
 +++ b/src/os/unix/ngx_process_cycle.c
 @@ -15,6 +15,8 @@ static void ngx_start_worker_processes(ngx_cycle_t *cycle, ngx_int_t n,
@@ -81,17 +69,7 @@ index 3ead164..6151e8a 100644
  
      ngx_new_binary = 0;
      delay = 0;
-@@ -188,7 +192,8 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
-                 continue;
-             }
- 
--            sigio = ccf->worker_processes + 2 /* cache processes */;
-+            sigio = ccf->worker_processes + 2   /* cache processes */
-+                                          + 1 ; /* privileged agent process */
- 
-             if (delay > 1000) {
-                 ngx_signal_worker_processes(cycle, SIGKILL);
-@@ -224,6 +229,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
+@@ -224,6 +228,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
                  ngx_start_worker_processes(cycle, ccf->worker_processes,
                                             NGX_PROCESS_RESPAWN);
                  ngx_start_cache_manager_processes(cycle, 0);
@@ -99,7 +77,7 @@ index 3ead164..6151e8a 100644
                  ngx_noaccepting = 0;
  
                  continue;
-@@ -243,6 +249,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
+@@ -243,6 +248,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
              ngx_start_worker_processes(cycle, ccf->worker_processes,
                                         NGX_PROCESS_JUST_RESPAWN);
              ngx_start_cache_manager_processes(cycle, 1);
@@ -107,7 +85,7 @@ index 3ead164..6151e8a 100644
  
              /* allow new processes to start */
              ngx_msleep(100);
-@@ -257,6 +264,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
+@@ -257,6 +263,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
              ngx_start_worker_processes(cycle, ccf->worker_processes,
                                         NGX_PROCESS_RESPAWN);
              ngx_start_cache_manager_processes(cycle, 0);
@@ -115,7 +93,7 @@ index 3ead164..6151e8a 100644
              live = 1;
          }
  
-@@ -424,6 +432,34 @@ ngx_start_cache_manager_processes(ngx_cycle_t *cycle, ngx_uint_t respawn)
+@@ -424,6 +431,34 @@ ngx_start_cache_manager_processes(ngx_cycle_t *cycle, ngx_uint_t respawn)
  
  
  static void
@@ -150,16 +128,20 @@ index 3ead164..6151e8a 100644
  ngx_pass_open_channel(ngx_cycle_t *cycle, ngx_channel_t *ch)
  {
      ngx_int_t  i;
-@@ -827,7 +863,7 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
+@@ -827,7 +862,11 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
          }
      }
  
 -    if (geteuid() == 0) {
-+    if (ngx_process != NGX_PROCESS_PRIVILEGED && geteuid() == 0) {
++    /*
++     * worker: -99
++     * privileged agent process have the same permission as master process
++     */
++    if (worker != -99 && geteuid() == 0) {
          if (setgid(ccf->group) == -1) {
              ngx_log_error(NGX_LOG_EMERG, cycle->log, ngx_errno,
                            "setgid(%d) failed", ccf->group);
-@@ -1144,6 +1180,43 @@ ngx_cache_manager_process_cycle(ngx_cycle_t *cycle, void *data)
+@@ -1144,6 +1183,47 @@ ngx_cache_manager_process_cycle(ngx_cycle_t *cycle, void *data)
  
  
  static void
@@ -167,19 +149,23 @@ index 3ead164..6151e8a 100644
 +{
 +    char              *name = data;
 +
++    ngx_setproctitle(name);
++
 +    /*
 +     * Set correct process type since closing listening Unix domain socket
 +     * in a master process also removes the Unix domain socket file.
 +     */
-+    ngx_process = NGX_PROCESS_PRIVILEGED;
++    ngx_process = NGX_PROCESS_HELPER;
 +
 +    ngx_close_listening_sockets(cycle);
 +
-+    ngx_worker_process_init(cycle, -1);
++    /*
++     * worker: -99
++     * privileged agent process have the same permission as master process
++     */
++    ngx_worker_process_init(cycle, -99);
 +
 +    ngx_use_accept_mutex = 0;
-+
-+    ngx_setproctitle(name);
 +
 +    for ( ;; ) {
 +
@@ -203,15 +189,3 @@ index 3ead164..6151e8a 100644
  ngx_cache_manager_process_handler(ngx_event_t *ev)
  {
      time_t        next, n;
-diff --git a/src/os/unix/ngx_process_cycle.h b/src/os/unix/ngx_process_cycle.h
-index 69495d5..eba7e8a 100644
---- a/src/os/unix/ngx_process_cycle.h
-+++ b/src/os/unix/ngx_process_cycle.h
-@@ -25,6 +25,7 @@
- #define NGX_PROCESS_SIGNALLER  2
- #define NGX_PROCESS_WORKER     3
- #define NGX_PROCESS_HELPER     4
-+#define NGX_PROCESS_PRIVILEGED 99
- 
- 
- typedef struct {
--- a/patches/nginx-1.11.2-privilege_process.patch
+++ b/patches/nginx-1.11.2-privilege_process.patch
@@ -1,46 +1,32 @@
 diff --git a/src/core/nginx.c b/src/core/nginx.c
-index 60f8fe7..b15f84c 100644
+index 60f8fe7..6989220 100644
 --- a/src/core/nginx.c
 +++ b/src/core/nginx.c
-@@ -53,6 +53,13 @@ static ngx_command_t  ngx_core_commands[] = {
-       offsetof(ngx_core_conf_t, master),
-       NULL },
- 
-+    { ngx_string("privilege_process"),
-+      NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_FLAG,
-+      ngx_conf_set_flag_slot,
-+      0,
-+      offsetof(ngx_core_conf_t, privilege),
-+      NULL },
-+
-     { ngx_string("timer_resolution"),
-       NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_TAKE1,
-       ngx_conf_set_msec_slot,
-@@ -981,6 +988,7 @@ ngx_core_module_create_conf(ngx_cycle_t *cycle)
+@@ -981,6 +981,7 @@ ngx_core_module_create_conf(ngx_cycle_t *cycle)
  
      ccf->daemon = NGX_CONF_UNSET;
      ccf->master = NGX_CONF_UNSET;
-+    ccf->privilege = NGX_CONF_UNSET;
++    ccf->privileged_worker = NGX_CONF_UNSET;
      ccf->timer_resolution = NGX_CONF_UNSET_MSEC;
  
      ccf->worker_processes = NGX_CONF_UNSET;
-@@ -1009,6 +1017,7 @@ ngx_core_module_init_conf(ngx_cycle_t *cycle, void *conf)
+@@ -1009,6 +1010,7 @@ ngx_core_module_init_conf(ngx_cycle_t *cycle, void *conf)
  
      ngx_conf_init_value(ccf->daemon, 1);
      ngx_conf_init_value(ccf->master, 1);
-+    ngx_conf_init_value(ccf->privilege, 0);
++    ngx_conf_init_value(ccf->privileged_worker, 0);
      ngx_conf_init_msec_value(ccf->timer_resolution, 0);
  
      ngx_conf_init_value(ccf->worker_processes, 1);
 diff --git a/src/core/ngx_cycle.h b/src/core/ngx_cycle.h
-index c51b7ff..6b396fe 100644
+index c51b7ff..740b528 100644
 --- a/src/core/ngx_cycle.h
 +++ b/src/core/ngx_cycle.h
 @@ -22,6 +22,9 @@
  #define NGX_DEBUG_POINTS_ABORT  2
  
  
-+#define HAVE_PRIVILEGE_PROCESS_PATCH 1
++#define HAVE_PRIVILEGED_PROCESS_PATCH 1
 +
 +
  typedef struct ngx_shm_zone_s  ngx_shm_zone_t;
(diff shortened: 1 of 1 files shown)


membphis:
missing macro define control if patched and test cases, I'll submit later.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
+diff --git a/src/core/nginx.c b/src/core/nginx.c
+index 60f8fe7..b15f84c 100644
+--- a/src/core/nginx.c
++++ b/src/core/nginx.c
+@@ -53,6 +53,13 @@ static ngx_command_t  ngx_core_commands[] = {
+       offsetof(ngx_core_conf_t, master),
+       NULL },
+ 
++    { ngx_string("privilege_process"),
```
agentzh: We don't need an explicit config directive to enable this. We only need a C API for 3rd-party module to request at run time. Just like the existing cache manager and cache loader processes have no explicit config directives, but are triggered automatically by the presence of any http cache directives like `proxy_cache`.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
++
++
+ typedef struct ngx_shm_zone_s  ngx_shm_zone_t;
+ 
+ typedef ngx_int_t (*ngx_shm_zone_init_pt) (ngx_shm_zone_t *zone, void *data);
+@@ -81,6 +84,7 @@ struct ngx_cycle_s {
+ typedef struct {
+     ngx_flag_t                daemon;
+     ngx_flag_t                master;
++    ngx_flag_t                privilege;
```
agentzh: Seems like we can just use `ccf->privileged_worker` field for the flag?

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
+             ngx_start_cache_manager_processes(cycle, 0);
++            ngx_start_privilege_worker_processes(cycle, 0);
+             live = 1;
+         }
+ 
+@@ -424,6 +435,34 @@ ngx_start_cache_manager_processes(ngx_cycle_t *cycle, ngx_uint_t respawn)
+ 
+ 
+ static void
++ngx_start_privilege_worker_processes(ngx_cycle_t *cycle, ngx_uint_t respawn)
```
agentzh: English: I think we should call it "privileged_worker` instead.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
++static void
+ ngx_pass_open_channel(ngx_cycle_t *cycle, ngx_channel_t *ch)
+ {
+     ngx_int_t  i;
+@@ -827,7 +866,7 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
+         }
+     }
+ 
+-    if (geteuid() == 0) {
++    if (ngx_process != NGX_PROCESS_PRIVILEGE && geteuid() == 0) {
```
agentzh: Why skip the privileged worker for the channel?

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
+ static void
++ngx_privilege_worker_process_cycle(ngx_cycle_t *cycle, void *data)
++{
++    ngx_privilege_worker_ctx_t *ctx = data;
++
++    /*
++     * Set correct process type since closing listening Unix domain socket
++     * in a master process also removes the Unix domain socket file.
++     */
++    ngx_process = NGX_PROCESS_PRIVILEGE;
```
agentzh: Again, we should use the word "privileged" instead.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
+ static void ngx_cache_manager_process_cycle(ngx_cycle_t *cycle, void *data);
++static void ngx_privilege_worker_process_cycle(ngx_cycle_t *cycle, void *data);
+ static void ngx_cache_manager_process_handler(ngx_event_t *ev);
+ static void ngx_cache_loader_process_handler(ngx_event_t *ev);
+ 
+@@ -63,6 +66,10 @@ static ngx_cache_manager_ctx_t  ngx_cache_loader_ctx = {
+     ngx_cache_loader_process_handler, "cache loader process", 60000
+ };
+ 
++static ngx_privilege_worker_ctx_t  ngx_privilege_worker_ctx = {
```
agentzh: This `ctx` construct looks strange to me. Why do we need?

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
+ 
+ 
+ typedef struct {
+@@ -34,6 +35,11 @@ typedef struct {
+ } ngx_cache_manager_ctx_t;
+ 
+ 
++typedef struct {
++    char                      *name;
++} ngx_privilege_worker_ctx_t;
```
agentzh: Can we get rid of this?

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,232 @@
...
+     time_t        next, n;
+diff --git a/src/os/unix/ngx_process_cycle.h b/src/os/unix/ngx_process_cycle.h
+index 69495d5..b8d97ba 100644
+--- a/src/os/unix/ngx_process_cycle.h
++++ b/src/os/unix/ngx_process_cycle.h
+@@ -25,6 +25,7 @@
+ #define NGX_PROCESS_SIGNALLER  2
+ #define NGX_PROCESS_WORKER     3
+ #define NGX_PROCESS_HELPER     4
++#define NGX_PROCESS_PRIVILEGE  5
```
agentzh: Can we just reuse the existing `NGX_PROCESS_HELPER` here?

patches/nginx-1.11.2-privilege_process.patch
membphis: good idea, its more flexible and we can easily control it at Lua land. How about `ngx.worker.privilege({enable=true})` ? and It only can be runned at `init_by_lua` phase.

patches/nginx-1.11.2-privilege_process.patch
membphis: just style copy from cache manager.

patches/nginx-1.11.2-privilege_process.patch
membphis: The logic inside is used to modify the process's privilege, the privileged process should have same privilege as master process, it don't need to do this.

patches/nginx-1.11.2-privilege_process.patch
membphis: we can't reuse the `NGX_PROCESS_HELPER`, the cache manager and cache loader will use `NGX_PROCESS_HELPER`. they were different process type.

patches/nginx-1.11.2-privilege_process.patch
membphis: ok, i'll remove it later.

patches/nginx-1.11.2-privilege_process.patch
agentzh: @membphis I think the name "privileged agent" is better.

membphis: Uniform naming to the "privileged agent process" :)

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,195 @@
...
+     ngx_conf_init_value(ccf->worker_processes, 1);
+diff --git a/src/core/ngx_cycle.h b/src/core/ngx_cycle.h
+index c51b7ff..1d38ab5 100644
+--- a/src/core/ngx_cycle.h
++++ b/src/core/ngx_cycle.h
+@@ -22,6 +22,9 @@
+ #define NGX_DEBUG_POINTS_ABORT  2
+ 
+ 
++#define HAVE_PRIVILEGED_PROCESS_PATCH 1
```
agentzh: I think the nginx coding style requires more than one space before the macro body.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,195 @@
...
+     time_t        next, n;
+diff --git a/src/os/unix/ngx_process_cycle.h b/src/os/unix/ngx_process_cycle.h
+index 69495d5..20701d3 100644
+--- a/src/os/unix/ngx_process_cycle.h
++++ b/src/os/unix/ngx_process_cycle.h
+@@ -25,6 +25,7 @@
+ #define NGX_PROCESS_SIGNALLER  2
+ #define NGX_PROCESS_WORKER     3
+ #define NGX_PROCESS_HELPER     4
++#define NGX_PROCESS_PRIVILEGED 5
```
agentzh: I suggest you use a very large number here, like 99, to avoid future conflicts when the mainstream nginx core introduces a new process type constant of the value 5.

agentzh:
@dndx Will you have a look at this? Thanks!

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,195 @@
+diff --git a/src/core/nginx.c b/src/core/nginx.c
+index 60f8fe7..4bd244b 100644
+--- a/src/core/nginx.c
++++ b/src/core/nginx.c
+@@ -981,6 +981,7 @@ ngx_core_module_create_conf(ngx_cycle_t *cycle)
+ 
+     ccf->daemon = NGX_CONF_UNSET;
+     ccf->master = NGX_CONF_UNSET;
++    ccf->privileged_agent = NGX_CONF_UNSET;
```
dndx: Could you point out where the configuration entry for `ccf->privileged_agent` is?

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,195 @@
...
+ static ngx_uint_t ngx_reap_children(ngx_cycle_t *cycle);
+@@ -24,6 +26,7 @@ static void ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker);
+ static void ngx_worker_process_exit(ngx_cycle_t *cycle);
+ static void ngx_channel_handler(ngx_event_t *ev);
+ static void ngx_cache_manager_process_cycle(ngx_cycle_t *cycle, void *data);
++static void ngx_privileged_agent_process_cycle(ngx_cycle_t *cycle, void *data);
+ static void ngx_cache_manager_process_handler(ngx_event_t *ev);
+ static void ngx_cache_loader_process_handler(ngx_event_t *ev);
+ 
+@@ -130,6 +133,7 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
```
dndx: Inside `ngx_master_process_cycle()`, we may have to set `sigio = ccf->worker_processes + 2 /* cache processes */;` to `sigio = ccf->worker_processes + 3 /* cache processes */;` instead. Otherwise worker processes may have less time to shutdown than intended.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,195 @@
...
+ static void
++ngx_privileged_agent_process_cycle(ngx_cycle_t *cycle, void *data)
++{
++    char              *name = data;
++
++    /*
++     * Set correct process type since closing listening Unix domain socket
++     * in a master process also removes the Unix domain socket file.
++     */
++    ngx_process = NGX_PROCESS_PRIVILEGED;
```
dndx: Here, instead of making a new type, I suggest we just use `NGX_PROCESS_HELPER` similiar to how cache manager works.

Otherwise signal handlers inside http://lxr.nginx.org/source/src/os/unix/ngx_process.c#0400 won't work.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,195 @@
...
++static void
+ ngx_pass_open_channel(ngx_cycle_t *cycle, ngx_channel_t *ch)
+ {
+     ngx_int_t  i;
+@@ -827,7 +862,7 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
+         }
+     }
+ 
+-    if (geteuid() == 0) {
++    if (ngx_process != NGX_PROCESS_PRIVILEGED && geteuid() == 0) {
```
dndx: Here, instead of using a new process type `NGX_PROCESS_PRIVILEGED`, we may also check to see if current process is privileged worker with checks like:

```
#define is_privileged_worker()   (ngx_processes[ngx_process_slot].proc == ngx_privileged_agent_process_cycle)
```

Again, `ngx_process` is being used in numerous places inside NGINX core and changing it is probably not the best idea.

patches/nginx-1.11.2-privilege_process.patch
membphis: to enable privileged agent, the 3rd c module could set `ccf->privileged_agent` to 1.

patches/nginx-1.11.2-privilege_process.patch
membphis: good catch. I'll submit new version later.

patches/nginx-1.11.2-privilege_process.patch
membphis: I tried this way right now, but I found it's more complicated than current version. 

`ngx_privileged_agent_process_cycle` is a static function, the macro only valid in `ngx_process_cycle.c`. 

the 3rd c module is not easy to get process type, we have to export a new variable or function if we reuse `NGX_PROCESS_HELPER`.

membphis: Thought for a while, I love to reuse `NGX_PROCESS_HELPER` if possible, how about we export a new variable named `ngx_sub_process`?

```
NGX_PROCESS_HELPER_CACHE_MANAGER    401
NGX_PROCESS_HELPER_CACHE_LOADER     402
NGX_PROCESS_HELPER_PRIVILEGED_AGENT 403
```

we can do more precise control if we need. `ngx_process` is the main process type, `ngx_sub_process` is the process sub type.

@dndx @agentzh

agentzh: @membphis I think all we need is just the ability to know if the current process is a privileged agent. We shouldn't worry about nginx's own helper processes at all. It's useless to us.

dndx: Agreed with @agentzh here.

dndx: @membphis What about:
```
#define is_privileged_worker()   (ngx_processes[ngx_process_slot].name == "privileged agent process")
```
since all statically allocated strings inside the same C program has the same address.

If that does not work, maybe try
```
#define is_privileged_worker()   (ngx_strcmp(ngx_processes[ngx_process_slot].name, "privileged agent process") == 0)
```
as well.

agentzh: @dndx The latter is better since it's more portable and thus more reliable.

membphis: ok, I'll submit new version later, although it's a little hacky :)

membphis: I tried but got failed. here is the reason :)

qq20170425-141452

dndx: @membphis You are right, looks like the above proposal does not quite work...

How about we do something even easier. The `worker` parameter of `ngx_worker_process_init()` is a signed integer and it apparently have effect when `>= 0`. I searched NGINX source code and appears that no one is using `worker < -1`. We can probably pass a magic `worker` parameter (e.g. -32767, defined with macro) to disable `setuid()`, while still exerting the same behavior as `worker = -1`.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,191 @@
...
+@@ -827,7 +862,11 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
+         }
+     }
+ 
+-    if (geteuid() == 0) {
++    /*
++     * worker: -99
++     * privileged agent process have the same permission as master process
++     */
++    if (worker != -99 && geteuid() == 0) {
```
agentzh: Oh, man, better give this magic number a name.

membphis:
@dndx @agentzh submit new version right now. marked the privileged agent process type with the global variable.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,211 @@
...
+ static void ngx_cache_manager_process_cycle(ngx_cycle_t *cycle, void *data);
++static void ngx_privileged_agent_process_cycle(ngx_cycle_t *cycle, void *data);
+ static void ngx_cache_manager_process_handler(ngx_event_t *ev);
+ static void ngx_cache_loader_process_handler(ngx_event_t *ev);
+ 
+@@ -51,6 +54,11 @@ sig_atomic_t  ngx_noaccept;
+ ngx_uint_t    ngx_noaccepting;
+ ngx_uint_t    ngx_restart;
+ 
++ngx_uint_t    ngx_is_privileged_agent;
```
dndx: Does not looks like this is useful since we already had `NGX_PROCESS_INIT_PRIVILEGED_AGENT`?

membphis: `NGX_PROCESS_INIT_PRIVILEGED_AGENT` is a macro-define, It is not necessary now, removed. :)

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,211 @@
...
++
++
++#define NGX_PROCESS_INIT_PRIVILEGED_AGENT     -99
```
dndx: Either this or `ngx_is_privileged_agent`, we don't need both. Also it appears that the current only does not actually uses `ngx_is_privileged_agent`.

patches/nginx-1.11.2-privilege_process.patch
membphis: sorry, git push failed. submitted right now.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,208 @@
...
+ ngx_pass_open_channel(ngx_cycle_t *cycle, ngx_channel_t *ch)
+ {
+     ngx_int_t  i;
+@@ -827,7 +864,10 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
+         }
+     }
+ 
+-    if (geteuid() == 0) {
++    /*
++     * privileged agent process have the same permission as master process
```
agentzh: grammar: "have" => "has"

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,208 @@
...
+         if (setgid(ccf->group) == -1) {
+             ngx_log_error(NGX_LOG_EMERG, cycle->log, ngx_errno,
+                           "setgid(%d) failed", ccf->group);
+@@ -1144,6 +1184,44 @@ ngx_cache_manager_process_cycle(ngx_cycle_t *cycle, void *data)
+ 
+ 
+ static void
++ngx_privileged_agent_process_cycle(ngx_cycle_t *cycle, void *data)
++{
++    char              *name = data;
```
agentzh: Too many spaces after `char`?

membphis: fixed. c9283d98aa6dd772ae9ab355eeaafc015378efda

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -155,7 +155,7 @@ index 3ead164..658adfc 100644
  static void
 +ngx_privileged_agent_process_cycle(ngx_cycle_t *cycle, void *data)
 +{
-+    char              *name = data;
++    char *name = data;
```
agentzh: I think we should at least use a tabful of spaces here ;) One space is too few.

patches/nginx-1.11.2-privilege_process.patch
```diff
@@ -0,0 +1,208 @@
+diff --git a/src/core/nginx.c b/src/core/nginx.c
```
agentzh: Hmm, seems like you forgot to apply this patch to the nginx core by editing `util/mirror-tarballs`.

agentzh: Also, I think the patch file name should be `nginx-1.11.2-privileged_agent_process.patch` instead.

Never mind, I'll rename it locally myself.

dndx:
Question: how do we tell whether the current worker is the privileged worker in Lua?

agentzh:
@dndx See https://github.com/openresty/lua-resty-core/pull/100

agentzh:
@membphis Merged with the following extra patch:

```diff
diff --git a/patches/nginx-1.11.2-privilege_process.patch b/patches/nginx-1.11.2-privileged_agent_process.patch
similarity index 99%
rename from patches/nginx-1.11.2-privilege_process.patch
rename to patches/nginx-1.11.2-privileged_agent_process.patch
index 734d597..0d8b21b 100644
--- a/patches/nginx-1.11.2-privilege_process.patch
+++ b/patches/nginx-1.11.2-privileged_agent_process.patch
@@ -155,7 +155,7 @@ index 3ead164..2b3f8df 100644
  static void
 +ngx_privileged_agent_process_cycle(ngx_cycle_t *cycle, void *data)
 +{
-+    char *name = data;
++    char   *name = data;
 +
 +    /*
 +     * Set correct process type since closing listening Unix domain socket
diff --git a/util/mirror-tarballs b/util/mirror-tarballs
index 2da01f1..263b5b4 100755
--- a/util/mirror-tarballs
+++ b/util/mirror-tarballs
@@ -55,6 +55,10 @@ else
     fi
 fi

+echo "$info_txt applying the privileged-agent-process patch"
+patch -p1 < $root/patches/nginx-$main_ver-privileged_agent_process.patch || exit 1
+echo
+
 echo "$info_txt applying the upstream-pipelining patch for nginx"
 patch -p1 < $root/patches/nginx-$main_ver-upstream_pipelining.patch || exit 1
 echo
diff --git a/util/ver b/util/ver
index 9aacdfe..82a0432 100755
--- a/util/ver
+++ b/util/ver
@@ -1,7 +1,7 @@
 #!/bin/bash

 main_ver=1.11.2
-minor_ver=3
+minor_ver=4rc0
 version=$main_ver.$minor_ver
 echo $version

```

Thanks!

PR #253, opened 2017-04-29
Also check that the building machine has sse4.2

xvaara:
Build fails on older machines that don't have sse4.2.
now we check the /proc/cpuinfo for the string sse4_2

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -638,7 +638,7 @@ _END_
             my $ofile = tmpnam();
             my $comp = ($cc || 'cc');
 
-            if (system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
+            if (system("grep -q sse4_2 /proc/cpuinfo") && system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
                 print "INFO: found -msse4.2 in $comp.\n";
                 $luajit_xcflags .= " -msse4.2";
                 unlink $ofile;
--- a/util/configure
+++ b/util/configure
@@ -628,21 +628,12 @@ _END_
 
         {
             # check -msse4.2
-            my ($out, $cfile) = tempfile("resty-config-XXXXXX",
-                                         SUFFIX => '.c', TMPDIR => 1,
-                                         UNLINK => 1);
-
-            print $out "int main(void) { return 0; }";
-            close $out;
-
-            my $ofile = tmpnam();
+            
             my $comp = ($cc || 'cc');
 
-            if (system("grep -q sse4_2 /proc/cpuinfo") && system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
+            if (system("echo '' |$comp -march=native -dM -E - |grep -q SSE4_2") == 0) {
                 print "INFO: found -msse4.2 in $comp.\n";
                 $luajit_xcflags .= " -msse4.2";
-                unlink $ofile;
-
             } else {
                 print "WARNING: -msse4.2 not supported in $comp.\n";
             }
--- a/util/configure
+++ b/util/configure
@@ -628,12 +628,28 @@ _END_
 
         {
             # check -msse4.2
-            
+            my ($out, $cfile) = tempfile("resty-config-XXXXXX",
+                                         SUFFIX => '.c', TMPDIR => 1,
+                                         UNLINK => 1);
+
+            print $out "
+int main(int argc, char** argv) {
+#if !defined(__SSE4_2__)
+#error not support
+#endif
+    return 0;
+}
+";
+            close $out;
+
+            my $ofile = tmpnam();
             my $comp = ($cc || 'cc');
 
-            if (system("echo '' |$comp -march=native -dM -E - |grep -q SSE4_2") == 0) {
+            if (system("$comp -o $ofile -march=native -c $cfile") == 0 && -s $ofile) {
                 print "INFO: found -msse4.2 in $comp.\n";
                 $luajit_xcflags .= " -msse4.2";
+                unlink $ofile;
+
             } else {
                 print "WARNING: -msse4.2 not supported in $comp.\n";
             }


util/configure
```diff
@@ -638,7 +638,7 @@ _END_
             my $ofile = tmpnam();
             my $comp = ($cc || 'cc');
 
-            if (system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
+            if (system("grep -q sse4_2 /proc/cpuinfo") && system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
```
agentzh: Hmm, `/proc/cpuinfo` is not portable. We need to support various different operating systems like Windows, FreeBSD, and Mac OS X. Maybe we should actually run the resulting C program to validate?

xvaara:
I don't have Windows, but that worked with my mac & linux, so I recon it would also work on freebsd.

util/configure
```diff
@@ -628,21 +628,12 @@ _END_
 
         {
             # check -msse4.2
-            my ($out, $cfile) = tempfile("resty-config-XXXXXX",
-                                         SUFFIX => '.c', TMPDIR => 1,
-                                         UNLINK => 1);
-
-            print $out "int main(void) { return 0; }";
```
agentzh: Can we actually use SSE 4.2 instructions in this test C program and actually runs the resulting binary to see if it returns 0 code?

xvaara: Tried to find something quickly when searching for solution. But I'm not a C-coder so I'm not comfortable doing that, since didn't find any easy way of doing that.

yangshuxin:
My $0.02 value of input to this patch:

1. "being portable" is a matter of degree, not a matter of black & white, and we are trying to improve the degree as further as possible.
2. examining "/proc/cpuinfo" is not very portable.
3. I could be wrong, but I kinda feel "-msse4.2" flag is more "portable" than the  `gcc -dM -march=native $* -E`. IIRC, quite a few compiler support the "-mssex.y" few years ago When I implement a vectorization about 6 years back. 
4. inline asm is not portable. gcc's inline asm is different from others, although many other static compilers try to keep in sync with gcc about inline asm.

xvaara:
@yangshuxin don't know about 3. The "-msse4.2" flag causes the build binaries to fail on a non sse4.2 machine. I know they are old, but still got some chucking along as test servers. The flag tests the compiler, not the machine.

[Example without any configure flags on the current tar.gz](https://gist.github.com/xvaara/27f2bdaafcfa49a93a5effacb0eb29d7)

> ~/src/openresty-1.11.2.3  ./build/luajit-root/usr/local/openresty/luajit/bin/luajit
> [1] 6082 illegal hardware instruction ./build/luajit-root/usr/local/openresty/luajit/bin/luajit

Put with the patch:
> ~/src/openresty-1.11.2.3  ./build/luajit-root/usr/local/openresty/luajit/bin/luajit
> LuaJIT 2.1.0-beta2 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
> JIT: ON SSE2 SSE3 SSE4.1 fold cse dce fwd dse narrow loop abc sink fuse

yangshuxin:
@xvaara , you are right as to  -mss4.2 flag in gcc does not check if the underlying architecture support sse4.2.

alternative, test can be done with this: 
```
int main(int argc, char** argv) {
#if !defined(__SSE4_2__)
#error not support
#endif
    return 0;
}
```
and compile with `-march=native`. I guess this is more portable than `gcc -dM -march=native $* -E`.

xvaara:
Seems to work, for me at least:
`cd LuaJIT-2.1-20170405
/tmp/resty-config-YHofzs.c: In function ‘main’:
/tmp/resty-config-YHofzs.c:3:2: error: #error not support
 #error not support
  ^
WARNING: -msse4.2 not supported in cc.`

yangshuxin:
The testing sse part LGTM.

xvaara:
Should we also add --without-sse42 option if someone needs it?

agentzh:
@xvaara I'd rather wait for someone who actually needs that :)

util/configure
```diff
@@ -632,13 +632,20 @@ _END_
...
+    return 0;
+}
+";
             close $out;
 
             my $ofile = tmpnam();
             my $comp = ($cc || 'cc');
 
-            if (system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
+            if (system("$comp -o $ofile -march=native -c $cfile") == 0 && -s $ofile) {
```
agentzh: This change is wrong. The whole point of this feature test is for the `-msse4.2` option.

agentzh:
@xvaara Merged with the following extra patch:

```diff
diff --git a/util/configure b/util/configure
index f6def03..e78fc74 100755
--- a/util/configure
+++ b/util/configure
@@ -633,9 +633,9 @@ _END_
                                          UNLINK => 1);

             print $out "
-int main(int argc, char** argv) {
-#if !defined(__SSE4_2__)
-#error not support
+int main(void) {
+#ifndef __SSE4_2__
+#   error SSE 4.2 not found
 #endif
     return 0;
 }
@@ -644,13 +644,23 @@ int main(int argc, char** argv) {

             my $ofile = tmpnam();
             my $comp = ($cc || 'cc');
+            my $found;

-            if (system("$comp -o $ofile -march=native -c $cfile") == 0 && -s $ofile) {
-                print "INFO: found -msse4.2 in $comp.\n";
-                $luajit_xcflags .= " -msse4.2";
+            if (system("$comp -o $ofile -msse4.2 -c $cfile") == 0 && -s $ofile) {
                 unlink $ofile;

-            } else {
+                if (system("$comp -o $ofile -march=native -c $cfile") == 0 && -s $ofile) {
+                    print "INFO: found -msse4.2 in $comp.\n";
+                    $found = 1;
+                    $luajit_xcflags .= " -msse4.2";
+                }
+            }
+
+            if (-f $ofile) {
+                unlink $ofile;
+            }
+
+            if (!$found) {
                 print "WARNING: -msse4.2 not supported in $comp.\n";
             }
         }
```

Let me know if it works for you. I don't have such an old CPU to test things.

Thanks!

xvaara:
It works. But to nitpick, the error message is wrong. Since sse4.2 is in the $comp, but not supported by the cpu. Minor difference. 

The --without-luajit-sse4.2 flag would be needed if you want to build debian packages, since debians own packages are build to work without sse4.2. Should luajit itself check for sse4.2 at runtime? Or does anyone anymore care about 9year old processors 😀

agentzh:
@xvaara For our own binary rpm/deb packages for OpenResty on x86_64, we do not want to support CPUs without SSE 4.2 support.

PR #266, opened 2017-06-16
optimization: reduce the number of ngx_connection_t allocated inside the

dndx:
privileged worker to avoid excessive memory consumption when
`worker_connections` is set very high.

When `worker_connections` is set to  `409600`. on x64 machine:

Before:
```
root      66953  0.0  8.9 519876 180556 ?       S    16:26   0:00 nginx: privileged agent process
```

After:
```
root      67955  0.0  0.5 350332 11100 ?        S    16:31   0:00 nginx: privileged agent process
```


agentzh:
@dndx Merged. Thanks!

PR #274, opened 2017-07-16
feature: make $proxy_add_x_forwarded_for accessible in Lua land.

spacewander:
This modification make `$proxy_add_x_forwarded_for` accessible via `ngx.var`, just like what have been done with `$proxy_host` and `$proxy_port`.
 
Patch originally discussed at https://github.com/openresty/lua-nginx-module/issues/1115

Looking forward for any feedback.


agentzh:
@spacewander Thanks for the patch. Will you please also add a test case to `openresty/lua-nginx-module` so that we can have this feature covered by our existing test suites? Thanks!

agentzh:
I've merged this for now.

PR #295, opened 2017-08-30
feature: added patched needed to support stream balancer_by_lua_*.

dndx:

### DIFF
--- a/patches/nginx-1.13.3-stream_balancer_export.patch
+++ b/patches/nginx-1.13.3-stream_balancer_export.patch
@@ -32,20 +32,22 @@ index 526de3a..b531ce1 100644
      void *data)
  {
 diff --git a/src/stream/ngx_stream_upstream_round_robin.h b/src/stream/ngx_stream_upstream_round_robin.h
-index 35d9fce..c721833 100644
+index 35d9fce..75f3e31 100644
 --- a/src/stream/ngx_stream_upstream_round_robin.h
 +++ b/src/stream/ngx_stream_upstream_round_robin.h
-@@ -141,6 +141,13 @@ ngx_int_t ngx_stream_upstream_get_round_robin_peer(ngx_peer_connection_t *pc,
-     void *data);
+@@ -142,5 +142,15 @@ ngx_int_t ngx_stream_upstream_get_round_robin_peer(ngx_peer_connection_t *pc,
  void ngx_stream_upstream_free_round_robin_peer(ngx_peer_connection_t *pc,
      void *data, ngx_uint_t state);
+ 
++#if (NGX_STREAM_SSL)
 +ngx_int_t ngx_stream_upstream_set_round_robin_peer_session(
 +    ngx_peer_connection_t *pc, void *data);
 +void ngx_stream_upstream_save_round_robin_peer_session(
 +    ngx_peer_connection_t *pc, void *data);
++#endif
 +
 +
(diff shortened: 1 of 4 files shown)


patches/nginx-1.13.3-stream_proxy_timeout_fields.patch
```diff
@@ -0,0 +1,168 @@
...
+ typedef ngx_int_t (*ngx_stream_filter_pt)(ngx_stream_session_t *s,
+diff --git a/src/stream/ngx_stream_proxy_module.c b/src/stream/ngx_stream_proxy_module.c
+index 0afde1c..f197c83 100644
+--- a/src/stream/ngx_stream_proxy_module.c
++++ b/src/stream/ngx_stream_proxy_module.c
+@@ -359,6 +359,7 @@ ngx_stream_proxy_handler(ngx_stream_session_t *s)
+     ngx_stream_proxy_srv_conf_t      *pscf;
+     ngx_stream_upstream_srv_conf_t   *uscf, **uscfp;
+     ngx_stream_upstream_main_conf_t  *umcf;
++    ngx_stream_proxy_ctx_t           *sctx;
```
agentzh: Why don't we just use `ctx` here?

patches/nginx-1.13.3-stream_proxy_timeout_fields.patch
```diff
@@ -0,0 +1,168 @@
...
++++ b/src/stream/ngx_stream.h
+@@ -241,6 +241,14 @@ typedef struct {
+ } ngx_stream_module_t;
+ 
+ 
++typedef struct {
++    ngx_msec_t                       connect_timeout;
++    ngx_msec_t                       timeout;
++} ngx_stream_proxy_ctx_t;
++
```
agentzh: Style: need 2 blank lines here.

patches/nginx-1.13.3-stream_balancer_export.patch
```diff
@@ -0,0 +1,51 @@
...
+--- a/src/stream/ngx_stream_upstream_round_robin.h
++++ b/src/stream/ngx_stream_upstream_round_robin.h
+@@ -141,6 +141,13 @@ ngx_int_t ngx_stream_upstream_get_round_robin_peer(ngx_peer_connection_t *pc,
+     void *data);
+ void ngx_stream_upstream_free_round_robin_peer(ngx_peer_connection_t *pc,
+     void *data, ngx_uint_t state);
++ngx_int_t ngx_stream_upstream_set_round_robin_peer_session(
++    ngx_peer_connection_t *pc, void *data);
++void ngx_stream_upstream_save_round_robin_peer_session(
++    ngx_peer_connection_t *pc, void *data);
```
agentzh: These 2 functions were originally protected by the macro `NGX_STREAM_SSL` and we should keep that in the header file as well.

patches/nginx-1.13.3-stream_proxy_timeout_fields.patch
dndx: `ngx_stream_proxy_handler` already defines `ctx` above.

agentzh: @dndx I see. Then we'd better name it to `pctx` instead of `sctx`.

agentzh:
@dndx Merged. Thanks!

PR #301, opened 2017-09-16
patches: ensure "server" header in HTTP/2 response shows "openresty"

dndx:
when server_tokens are turned off.


agentzh:
@dndx I think we just need the patch for nginx 1.13.5, no earlier versions needed.

dndx:
@agentzh Added. I think we can just keep the changes to older versions as they are not harmful anyway.

agentzh:
@dndx Fine.

agentzh:
@dndx Seems like you forgot to update `util/mirror-tarballs` to actually apply the patch?

dndx:
@agentzh We are updating an existing patch so no need to change `util/mirror-tarballs.

agentzh:
@dndx Merged. Thanks!

PR #342, opened 2018-02-09
feature: added the socket_cloexec patch to ensure connection could be closed before child process terminates

spacewander:

### DIFF
--- a/patches/nginx-1.13.6-socket_cloexec.patch
+++ b/patches/nginx-1.13.6-socket_cloexec.patch
@@ -92,29 +92,33 @@ index c5bb8068..b4920655 100644
      ngx_log_debug2(NGX_LOG_DEBUG_EVENT, pc->log, 0, "%s socket %d",
                     (type == SOCK_STREAM) ? "stream" : "dgram", s);
 diff --git a/src/os/unix/ngx_socket.h b/src/os/unix/ngx_socket.h
-index fcc51533..1faf5a00 100644
+index fcc51533..7dda965c 100644
 --- a/src/os/unix/ngx_socket.h
 +++ b/src/os/unix/ngx_socket.h
-@@ -38,6 +38,9 @@ int ngx_blocking(ngx_socket_t s);
+@@ -38,6 +38,11 @@ int ngx_blocking(ngx_socket_t s);
  
  #endif
  
 +#define ngx_cloexec(s)      fcntl(s, F_SETFD, FD_CLOEXEC)
 +#define ngx_cloexec_n(s)    "fcntl(FD_CLOEXEC)"
 +
++#define HAVE_SOCKET_CLOEXEC_PATCH  1
++
  int ngx_tcp_nopush(ngx_socket_t s);
  int ngx_tcp_push(ngx_socket_t s);
  
 diff --git a/src/os/win32/ngx_socket.h b/src/os/win32/ngx_socket.h
-index a9e26c29..14113d96 100644
+index a9e26c29..fb5b622a 100644
 --- a/src/os/win32/ngx_socket.h
 +++ b/src/os/win32/ngx_socket.h
-@@ -31,6 +31,9 @@ int ngx_blocking(ngx_socket_t s);
+@@ -31,6 +31,11 @@ int ngx_blocking(ngx_socket_t s);
  #define ngx_nonblocking_n   "ioctlsocket(FIONBIO)"
  #define ngx_blocking_n      "ioctlsocket(!FIONBIO)"
  
 +#define ngx_cloexec(s)      1
 +#define ngx_cloexec_n(s)    ""
++
++#define HAVE_SOCKET_CLOEXEC_PATCH  1
 +
  #define ngx_shutdown_socket    shutdown
  #define ngx_shutdown_socket_n  "shutdown()"
--- a/patches/nginx-1.13.6-socket_cloexec.patch
+++ b/patches/nginx-1.13.6-socket_cloexec.patch
@@ -1,5 +1,5 @@
 diff --git a/auto/unix b/auto/unix
-index 10835f6c..16ea377b 100644
+index 10835f6c..78a8073a 100644
 --- a/auto/unix
 +++ b/auto/unix
 @@ -990,3 +990,14 @@ ngx_feature_test='struct addrinfo *res;
@@ -15,7 +15,7 @@ index 10835f6c..16ea377b 100644
 +ngx_feature_path=
 +ngx_feature_libs=
 +ngx_feature_test="int fd;
-+                  fd = socket(AF_INET, SOCK_STREAM, SOCK_CLOEXEC);
++                  fd = socket(AF_INET, SOCK_STREAM, SOCK_CLOEXEC);"
 +. auto/feature
 diff --git a/src/core/ngx_connection.c b/src/core/ngx_connection.c
 index 9a747589..56a8a955 100644
@@ -92,7 +92,7 @@ index c5bb8068..b4920655 100644
      ngx_log_debug2(NGX_LOG_DEBUG_EVENT, pc->log, 0, "%s socket %d",
                     (type == SOCK_STREAM) ? "stream" : "dgram", s);
 diff --git a/src/os/unix/ngx_socket.h b/src/os/unix/ngx_socket.h
-index fcc51533..7dda965c 100644
+index fcc51533..74f90cc6 100644
 --- a/src/os/unix/ngx_socket.h
 +++ b/src/os/unix/ngx_socket.h
 @@ -38,6 +38,11 @@ int ngx_blocking(ngx_socket_t s);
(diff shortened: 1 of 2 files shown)


patches/nginx-1.13.6-socket_cloexec.patch
```diff
@@ -0,0 +1,121 @@
...
+diff --git a/src/os/win32/ngx_socket.h b/src/os/win32/ngx_socket.h
+index a9e26c29..14113d96 100644
+--- a/src/os/win32/ngx_socket.h
++++ b/src/os/win32/ngx_socket.h
+@@ -31,6 +31,9 @@ int ngx_blocking(ngx_socket_t s);
+ #define ngx_nonblocking_n   "ioctlsocket(FIONBIO)"
+ #define ngx_blocking_n      "ioctlsocket(!FIONBIO)"
+ 
++#define ngx_cloexec(s)      1
++#define ngx_cloexec_n(s)    ""
```
dndx: This should be `#define ngx_cloexec_n    "ngx_cloexec()"` instead?

spacewander: Well, there is not a real function called `ngx_cloexec`, so using it may be confusing. `xxx_n` macro will be expanded to the actual syscall, and the actual syscall of `ngx_cloexec` under win32 is nothing.

agentzh:
@spacewander I tried to build the openresty produced from the branch of this PR on my Fedora 27 x64 but it failed horribly:

```
auto/unix: line 1001: unexpected EOF while looking for matching `"'
auto/unix: line 1004: syntax error: unexpected end of file
In file included from src/core/ngx_core.h:58:0,
                 from src/core/ngx_connection.c:9:
src/core/ngx_connection.c: In function ‘ngx_open_listening_sockets’:
src/core/ngx_connection.c:461:31: error: ‘ngx_cloexec_n’ undeclared (first use in this function); did you mean ‘ngx_iovec_t’?
                               ngx_cloexec_n " %V failed", &ls[i].addr_text);
                               ^
src/core/ngx_log.h:88:67: note: in definition of macro ‘ngx_log_error’
     if ((log)->log_level >= level) ngx_log_error_core(level, log, __VA_ARGS__)
                                                                   ^~~~~~~~~~~
src/core/ngx_connection.c:461:31: note: each undeclared identifier is reported only once for each function it appears in
                               ngx_cloexec_n " %V failed", &ls[i].addr_text);
                               ^
src/core/ngx_log.h:88:67: note: in definition of macro ‘ngx_log_error’
     if ((log)->log_level >= level) ngx_log_error_core(level, log, __VA_ARGS__)
                                                                   ^~~~~~~~~~~
src/core/ngx_connection.c:461:45: error: expected ‘)’ before string constant
                               ngx_cloexec_n " %V failed", &ls[i].addr_text);
                                             ^
src/core/ngx_log.h:88:67: note: in definition of macro ‘ngx_log_error’
     if ((log)->log_level >= level) ngx_log_error_core(level, log, __VA_ARGS__)
                                                                   ^~~~~~~~~~~
make[2]: *** [objs/Makefile:938: objs/src/core/ngx_connection.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [Makefile:8: build] Error 2
make: *** [Makefile:8: all] Error 2
Command "make -j9 > /dev/null" failed at ../util/dist-check line 255, <DATA> line 1.
```

spacewander:
@agentzh 
Fixed in the latest commit.

patches/nginx-1.13.6-socket_cloexec.patch
```diff
@@ -0,0 +1,125 @@
...
+--- a/src/event/ngx_event_accept.c
++++ b/src/event/ngx_event_accept.c
+@@ -62,7 +62,8 @@ ngx_event_accept(ngx_event_t *ev)
+ 
+ #if (NGX_HAVE_ACCEPT4)
+         if (use_accept4) {
+-            s = accept4(lc->fd, &sa.sockaddr, &socklen, SOCK_NONBLOCK);
++            s = accept4(lc->fd, &sa.sockaddr, &socklen,
++                        SOCK_NONBLOCK | SOCK_CLOEXEC);
+         } else {
```
agentzh: Style: needs a blank line before `} else {`.

patches/nginx-1.13.6-socket_cloexec.patch
```diff
@@ -0,0 +1,125 @@
...
++ngx_feature="SOCK_CLOEXEC support"
++ngx_feature_name="NGX_HAVE_SOCKET_CLOEXEC"
++ngx_feature_run=no
++ngx_feature_incs="#include <sys/types.h>
++                  #include <sys/socket.h>"
++ngx_feature_path=
++ngx_feature_libs=
++ngx_feature_test="int fd;
++                  fd = socket(AF_INET, SOCK_STREAM, SOCK_CLOEXEC);"
++. auto/feature
```
agentzh: It's better to also test the availability of `fcntl(s, F_SETFD, FD_CLOEXEC)` here?

patches/nginx-1.13.6-socket_cloexec.patch
```diff
@@ -0,0 +1,127 @@
...
++ngx_feature_name="NGX_HAVE_SOCKET_CLOEXEC"
++ngx_feature_run=no
++ngx_feature_incs="#include <sys/types.h>
++                  #include <sys/socket.h>
++                  #include <fcntl.h>"
++ngx_feature_path=
++ngx_feature_libs=
++ngx_feature_test="int fd;
++                  fd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0);
++                  fcntl(fd, F_SETFD, FD_CLOEXEC);"
```
agentzh: Seems like you misunderstood me :)

It's better to make these 2 separate feature tests. They are different things. So that maybe we can remove the win32 special handling and it will also automatically work on other systems lacking support for even the `FD_CLOEXEC` thing. It's safer and more generic.

patches/nginx-1.13.6-socket_cloexec.patch
```diff
@@ -0,0 +1,127 @@
...
++                        SOCK_NONBLOCK | SOCK_CLOEXEC);
+         } else {
+             s = accept(lc->fd, &sa.sockaddr, &socklen);
+         }
+@@ -202,6 +203,13 @@ ngx_event_accept(ngx_event_t *ev)
+                     ngx_close_accepted_connection(c);
+                     return;
+                 }
++
++                if (ngx_cloexec(s) == -1) {
```
agentzh: This call should be protected by a feature test macro also.

patches/nginx-1.13.6-socket_cloexec.patch
agentzh: Missing this one?

spacewander:
@agentzh
Fixed.

patches/nginx-1.13.6-socket_cloexec.patch
```diff
@@ -0,0 +1,142 @@
...
++++ b/src/os/unix/ngx_socket.h
+@@ -38,6 +38,17 @@ int ngx_blocking(ngx_socket_t s);
+ 
+ #endif
+ 
++#if (NGX_HAVE_FD_CLOEXEC)
++
++#define ngx_cloexec(s)      fcntl(s, F_SETFD, FD_CLOEXEC)
++#define ngx_cloexec_n       "fcntl(FD_CLOEXEC)"
++
```
agentzh: Better put the `#endif` here?

spacewander: @agentzh 
If `NGX_HAVE_FD_CLOEXEC` is not supported, this patch actually does nothing.
Maybe we should rename `HAVE_SOCKET_CLOEXEC_PATCH` to `ENABLE_SOCKET_CLOEXEC` to make it clear?

agentzh: @spacewander I think that macro is only an indicator for the existence of the patch itself, instead of indicating the existence of the actual feature :)

But I agree `HAVE_SOCKET_CLOEXEC_PATCH` is a bit confusing and should be renamed to `HAVE_CLOEXEC_PATCH` instead (`ENABLE_SOCKET_CLOEXEC` is actually worse, I'm afraid, which breaks our macro naming convention for such nginx core patches).

dndx:
There seems to be [some other files](http://lxr.nginx.org/ident?_i=ngx_socket&_remember=1) that creates sockets in additional to the ones that have been patched. Like `/src/core/ngx_syslog.c`, `/src/core/ngx_resolver.c`. Any thoughts on those ones?

spacewander:
@dndx 
Thank you for your remind. I just add the same process to `core/ngx_resolver.c@ngx_tcp_connect`
This patch's purpose is avoid holding connection until child process terminated, so just handle TCP socket is enough.

PR #348, opened 2018-02-24
feature: use resolv.conf from local machine as core resolvers

dndx:

Usage:
**resolver** address ... [valid=time] [ipv6=on|off] [local=on|off|path];

This can be used inside both `http` and `stream` subsystem.

If `local=on`, then in addition to the resolvers specified, NGINX will parse `/etc/resolv.conf` and use nameservers found inside as resolvers. You can also specify `local=path` to use a different file instead. By default, the path is relative to `nginx_prefix/conf`.

`local=off` will disable the parsing (this is the default).

Note that we only parse the first 2048 bytes of data from the `resolve.conf` file. In practice this should be more than enough as most of the time the size of this file is well under few hundred bytes.

### DIFF
--- a/patches/nginx-1.13.6-resolver_conf_parsing.patch
+++ b/patches/nginx-1.13.6-resolver_conf_parsing.patch
@@ -1,5 +1,5 @@
 diff --git a/src/core/ngx_resolver.c b/src/core/ngx_resolver.c
-index cd55520c..9156fadd 100644
+index cd55520c..d8dc49e8 100644
 --- a/src/core/ngx_resolver.c
 +++ b/src/core/ngx_resolver.c
 @@ -5,6 +5,7 @@
@@ -39,7 +39,7 @@ index cd55520c..9156fadd 100644
 +    ngx_file_t                       file;
 +    u_char                           buf[NGX_RESOLVER_FILE_BUF_SIZE];
 +    u_char                           ipv6_buf[NGX_INET6_ADDRSTRLEN];
-+    ngx_uint_t                       address, j, total = 0;
++    ngx_uint_t                       address = 0, j, total = 0;
 +    ssize_t                          n, i;
 +    enum {
 +        sw_nameserver,
--- a/patches/nginx-1.13.6-resolver_conf_parsing.patch
+++ b/patches/nginx-1.13.6-resolver_conf_parsing.patch
@@ -1,5 +1,5 @@
 diff --git a/src/core/ngx_resolver.c b/src/core/ngx_resolver.c
-index cd55520c..6a3f98bb 100644
+index cd55520c..d91c406c 100644
 --- a/src/core/ngx_resolver.c
 +++ b/src/core/ngx_resolver.c
 @@ -5,6 +5,7 @@
@@ -25,28 +25,31 @@ index cd55520c..6a3f98bb 100644
  
  typedef struct {
      u_char  ident_hi;
-@@ -131,6 +140,138 @@ static ngx_resolver_node_t *ngx_resolver_lookup_addr6(ngx_resolver_t *r,
+@@ -131,6 +140,180 @@ static ngx_resolver_node_t *ngx_resolver_lookup_addr6(ngx_resolver_t *r,
  #endif
  
  
 +static ngx_int_t
-+ngx_resolver_read_resolv_conf(ngx_conf_t *cf, ngx_resolver_t *r, u_char *path)
++ngx_resolver_read_resolv_conf(ngx_conf_t *cf, ngx_resolver_t *r, u_char *path,
++    size_t path_len)
 +{
 +    ngx_url_t                        u;
 +    ngx_resolver_connection_t       *rec;
 +    ngx_fd_t                         fd;
 +    ngx_file_t                       file;
 +    u_char                           buf[NGX_RESOLVER_FILE_BUF_SIZE];
++    u_char                           ipv6_buf[NGX_INET6_ADDRSTRLEN];
 +    ngx_uint_t                       address, j, total = 0;
 +    ssize_t                          n, i;
 +    enum {
 +        sw_nameserver,
++        sw_spaces,
 +        sw_address,
 +        sw_skip
 +    } state;
 +
 +    file.name.data = path;
-+    file.name.len = ngx_strlen(path);
++    file.name.len = path_len;
 +
 +    if (ngx_conf_full_name(cf->cycle, &file.name, 1) != NGX_OK) {
 +        return NGX_ERROR;
@@ -71,6 +74,11 @@ index cd55520c..6a3f98bb 100644
 +
 +    n = ngx_read_file(&file, buf, NGX_RESOLVER_FILE_BUF_SIZE, 0);
 +
++    if (n == NGX_ERROR) {
++        ngx_conf_log_error(NGX_LOG_ALERT, cf, ngx_errno,
++                           ngx_read_file_n " \"%s\" failed", file.name.data);
++    }
++
 +    if (ngx_close_file(file.fd) == NGX_FILE_ERROR) {
 +        ngx_conf_log_error(NGX_LOG_ALERT, cf, ngx_errno,
 +                           ngx_close_file_n " \"%s\" failed", file.name.data);
@@ -93,15 +101,22 @@ index cd55520c..6a3f98bb 100644
 +
 +        case sw_nameserver:
 +
-+            if ((size_t) n - i >= sizeof("nameserver ") - 1
-+                && ngx_memcmp(buf + i, "nameserver ",
-+                              sizeof("nameserver ") - 1) == 0)
++            if ((size_t) n - i >= sizeof("nameserver") - 1
++                && ngx_memcmp(buf + i, "nameserver",
++                              sizeof("nameserver") - 1) == 0)
 +            {
-+                state = sw_address;
-+                i += sizeof("nameserver ") - 1;
-+                address = i;
++                state = sw_spaces;
++                i += sizeof("nameserver") - 1;
 +
 +                continue;
++            }
++
++            break;
++
++        case sw_spaces:
++            if (buf[i] != '\t' && buf[i] != ' ') {
++                address = i;
++                state = sw_address;
 +            }
 +
 +            break;
@@ -112,8 +127,26 @@ index cd55520c..6a3f98bb 100644
 +                ngx_memzero(&u, sizeof(ngx_url_t));
 +
 +                u.url.data = buf + address;
-+                u.url.len = i - address;
++                u.url.len = i == n - 1 ? n - address : i - address;
 +                u.default_port = 53;
++
++                /* IPv6? */
++                if (ngx_strlchr(u.url.data, u.url.data + u.url.len, ':') != NULL) {
++                    if (u.url.len + 2 > sizeof(ipv6_buf)) {
++                        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
++                                           "IPv6 resolver address is too long: \"%V\"",
++                                           &u.url);
++
++                        return NGX_ERROR;
++                    }
++
++                    ipv6_buf[0] = '[';
++                    ngx_memcpy(ipv6_buf + 1, u.url.data, u.url.len);
++                    ipv6_buf[u.url.len + 1] = ']';
++
++                    u.url.data = ipv6_buf;
++                    u.url.len = u.url.len + 2;
++                }
 +
 +                if (ngx_parse_url(cf->pool, &u) != NGX_OK) {
 +                    if (u.err) {
@@ -141,6 +174,15 @@ index cd55520c..6a3f98bb 100644
 +
 +                total++;
 +
++#if (NGX_DEBUG)
++                /*
++                 * logs with level below NGX_LOG_NOTICE will not be printed
++                 * in this early phase
++                 */
++                ngx_conf_log_error(NGX_LOG_NOTICE, cf, 0,
++                                   "parsed a resolver: \"%V\"", &u.url);
++#endif
++
 +                state = sw_nameserver;
 +            }
 +
@@ -164,7 +206,7 @@ index cd55520c..6a3f98bb 100644
  ngx_resolver_t *
  ngx_resolver_create(ngx_conf_t *cf, ngx_str_t *names, ngx_uint_t n)
  {
-@@ -246,6 +387,33 @@ ngx_resolver_create(ngx_conf_t *cf, ngx_str_t *names, ngx_uint_t n)
+@@ -246,6 +429,35 @@ ngx_resolver_create(ngx_conf_t *cf, ngx_str_t *names, ngx_uint_t n)
          }
  #endif
  
@@ -172,7 +214,8 @@ index cd55520c..6a3f98bb 100644
 +
 +            if (ngx_strcmp(&names[i].data[6], "on") == 0) {
 +                if (ngx_resolver_read_resolv_conf(cf, r,
-+                                                  (u_char *) NGX_RESOLVER_FILE_NAME)
++                                                  (u_char *) NGX_RESOLVER_FILE_NAME,
++                                                  sizeof(NGX_RESOLVER_FILE_NAME) - 1)
 +                    != NGX_OK)
 +                {
 +                    ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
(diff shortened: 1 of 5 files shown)


patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,192 @@
+diff --git a/src/core/ngx_resolver.c b/src/core/ngx_resolver.c
+index cd55520c..06d1c461 100644
+--- a/src/core/ngx_resolver.c
++++ b/src/core/ngx_resolver.c
+@@ -15,6 +15,9 @@
+ #define NGX_RESOLVER_TCP_RSIZE  (2 + 65535)
+ #define NGX_RESOLVER_TCP_WSIZE  8192
+ 
++#define NGX_RESOLVER_FILE_BUF_SIZE  2048
```
thibaultcha: If we follow the approach taken in #247 (taken from other parts of the nginx codebase), we can easily parse files that exceed this size, if desired.

https://github.com/openresty/openresty/pull/247/files#diff-1f78491f86d1e1677eca130eb1799f68R198

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,192 @@
...
++                }
++
++                address = 0;
++                state = sw_nameserver;
++            }
++
++            break;
++
++        case sw_skip:
++            if (buf[i] == LF) {
```
thibaultcha: Should we also test against `CR`? Other file parsers from the nginx codebase do

https://github.com/nginx/nginx/blob/0ad556fe59ad132dc4d34dea9e80f2ff2c3c1314/src/core/ngx_conf_file.c#L752

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,192 @@
...
++                    if (u.err) {
++                        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
++                                           "%s in resolver \"%V\"",
++                                           u.err, &u.url);
++                    }
++
++                    return NGX_ERROR;
++                }
++
++                rec = ngx_array_push_n(&r->connections, u.naddrs);
```
thibaultcha: In hindsight, one of the things that I believe #247 is missing is the lack of test for whether a given IP address is already cached by the resolver or not.

Similarly in this case, do we care/are we doing anything to prevent the same nameserver from being added twice? Like:
```
# /etc/resolv.conf
nameserver 8.8.8.8

# nginx.conf
http {
    resolver 8.8.8.8;
    resolver local=on;
}
```

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,192 @@
...
++                ngx_memzero(rec, u.naddrs * sizeof(ngx_resolver_connection_t));
++
++                for (j = 0; j < u.naddrs; j++) {
++                    rec[j].sockaddr = u.addrs[j].sockaddr;
++                    rec[j].socklen = u.addrs[j].socklen;
++                    rec[j].server = u.addrs[j].name;
++                    rec[j].resolver = r;
++                }
++
++                address = 0;
```
thibaultcha: I think we might not need this instruction? Since the `sw_nameserver` state will take care of if before entering the `sw_address` state?

patches/nginx-1.13.6-resolver_conf_parsing.patch
dndx: I think testing against `LF` is sufficient as all modern *nix O/S uses `LF` as newline and the [ones that uses `CR` alone](https://en.wikipedia.org/wiki/Newline#Representations_in_different_character_encoding_specifications) are all legacy systems that no one still uses. Windows does uses `CRLF` but I don't think they have `resolv.conf` anyway..

patches/nginx-1.13.6-resolver_conf_parsing.patch
dndx: That's interesting. AFAIK [you can't do multiple `resolver` directives within the same block](http://lxr.nginx.org/source/src/http/ngx_http_core_module.c#5112). However, NGINX does not prohibit you from doing:

```
resolver 8.8.8.8 8.8.8.8;
```

either. In the description of this PR, I described the behavior of `local` as:

> If local=on, then in addition to the resolvers specified, NGINX will parse /etc/resolv.conf and use nameservers found inside as resolvers.

It seems that allowing duplicates makes more sense? Plus de-duping this will be very painful due to how those data structure works currently.

patches/nginx-1.13.6-resolver_conf_parsing.patch
dndx: Actually second thought, maybe supporting `CRLF` in addition to `LF` would be a better idea. However, I still think `CR` alone makes little sense in this case.

patches/nginx-1.13.6-resolver_conf_parsing.patch
thibaultcha: Yeah, sorry I think my example should have been:
```
# nginx.conf
http {
    resolver 8.8.8.8 local=on;
}
```

> It seems that allowing duplicates makes more sense? Plus de-duping this will be very painful due to how those data structure works currently.

Agreed - just raising the concern for us to be aware, but if nginx allows it, it is probably better to not try to add too much magic around it.

thibaultcha:
Do we plan on supporting `MAXNS`? FYI, the current maximum number of allowed nameserver entries  in `/etc/resolv.conf` is 3, as per:

http://man7.org/linux/man-pages/man5/resolv.conf.5.html

In the past, we rolled out our own `/etc/resolv.conf` parser (in Lua) without supporting this limit, and this caused a wild variety of issues among our customers who had more than 3 nameserver entries. The tail end of that list was never used by any of their other applications, but when we started parsing and using those nameservers, we caused them so issues since those entries were long stale and not working anymore (and they obviously never noticed before).

dndx:
@thibaultcha Yes I think that is a good idea. We should definitely keep the semantics similar to that of Glibc to avoid surprises.

patches/nginx-1.13.6-resolver_conf_parsing.patch
dndx: I think 2KB is more than enough for all non-abusing users. It's never a good idea to read the whole file as there is a possibility of the filesystem being broken or /etc/resolv.conf being symlinked to `/dev/zero`.

My observations is that `resolv.conf` will rarely exceed 500 bytes even with crazy comments and so on, so making it larger than 2KB seems to be unnecessarily large.

dndx:
@spacewander Could you please help review this PR as well so we can get it merged soon. Thanks!

dndx:
@thibaultcha We now respect `MAXNS` as well as line feed of `CR`, `LF` or `CRLF` as well.

thibaultcha:
@dndx Nice :+1:

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,201 @@
...
++
++    if (n == NGX_ERROR) {
++        return NGX_ERROR;
++    }
++
++    if (n == 0) {
++        return NGX_OK;
++    }
++
++    for (i = 0; i < n; total < MAXNS) {
```
spacewander: Do you mean `i < n && total < MAXNS`?

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,201 @@
...
+ 
++#include <resolv.h>
+ #include <ngx_config.h>
+ #include <ngx_core.h>
+ #include <ngx_event.h>
+@@ -15,6 +16,9 @@
+ #define NGX_RESOLVER_TCP_RSIZE  (2 + 65535)
+ #define NGX_RESOLVER_TCP_WSIZE  8192
+ 
++#define NGX_RESOLVER_FILE_BUF_SIZE  2048
```
spacewander: Note that there is not limitation for the resolv.conf file size. You could even writte poem in the comment! It would be better to put some comment here to mention that `2048` could cover most of cases.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,201 @@
...
++                           ngx_open_file_n " \"%s\" failed", file.name.data);
++
++        return NGX_ERROR;
++    }
++
++    ngx_memzero(&file, sizeof(ngx_file_t));
++
++    file.fd = fd;
++    file.name.data = path;
++    file.name.len = ngx_strlen(path) - 1;
```
spacewander: Are those lines duplicate? `file.name` has been assigned above.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,201 @@
...
++
++                continue;
++            }
++
++            break;
++
++        case sw_address:
++
++            if (address == 0) {
++                address = i;
```
spacewander: I think `address` could be directly assigned in the `sw_nameserver` status?

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,201 @@
...
+ {
+@@ -246,6 +388,33 @@ ngx_resolver_create(ngx_conf_t *cf, ngx_str_t *names, ngx_uint_t n)
+         }
+ #endif
+ 
++        if (ngx_strncmp(names[i].data, "local=", 6) == 0) {
++
++            if (ngx_strcmp(&names[i].data[6], "on") == 0) {
++                if (ngx_resolver_read_resolv_conf(cf, r,
++                                                  NGX_RESOLVER_FILE_NAME)
```
spacewander: Better to add `(u_char *) ` type cast to eliminate this warning:
```
src/core/ngx_resolver.c:398:51: warning: passing 'char [17]' to parameter of type 'u_char *' (aka 'unsigned char *') converts between pointers to integer types with different sign
      [-Wpointer-sign]
                                                  NGX_RESOLVER_FILE_NAME)
                                                  ^~~~~~~~~~~~~~~~~~~~~~
src/core/ngx_resolver.c:20:37: note: expanded from macro 'NGX_RESOLVER_FILE_NAME'
#define NGX_RESOLVER_FILE_NAME      "/etc/resolv.conf"
                                    ^~~~~~~~~~~~~~~~~~
src/core/ngx_resolver.c:139:74: note: passing argument to parameter 'path' here
ngx_resolver_read_resolv_conf(ngx_conf_t *cf, ngx_resolver_t *r, u_char *path
```

patches/nginx-1.13.6-resolver_conf_parsing.patch
dndx: Actually file.name would have changed after the `ngx_conf_full_name` call but seems to me that we don't need to assign it back anyway (`ngx_conf_full_name` also allocates NULL terminated string). I removed those two lines.

patches/nginx-1.13.6-resolver_conf_parsing.patch
dndx: Good catch, that was a big mistake, fixed.

dndx:
@spacewander Changes has been pushed. Please check again. Thanks!

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,200 @@
...
++    ngx_uint_t                       address, j, total = 0;
++    ssize_t                          n, i;
++    enum {
++        sw_nameserver,
++        sw_address,
++        sw_skip
++    } state;
++
++    file.name.data = path;
++    file.name.len = ngx_strlen(path);
```
agentzh: Can we avoid this `ngx_strlen()` call? It seems that the caller already knows the string length but you simply discards it.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,200 @@
...
++
++    n = ngx_read_file(&file, buf, NGX_RESOLVER_FILE_BUF_SIZE, 0);
++
++    if (ngx_close_file(file.fd) == NGX_FILE_ERROR) {
++        ngx_conf_log_error(NGX_LOG_ALERT, cf, ngx_errno,
++                           ngx_close_file_n " \"%s\" failed", file.name.data);
++    }
++
++    if (n == NGX_ERROR) {
++        return NGX_ERROR;
```
agentzh: It's unfortunate that you discard the errno (actually the errno may already get overridden by the earlier `ngx_close_file` call. We should log the system error message for failed `ngx_read_file()` calls.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -0,0 +1,200 @@
...
++    for (i = 0; i < n && total < MAXNS; /* void */) {
++        if (buf[i] == '#' || buf[i] == ';') {
++            state = sw_skip;
++        }
++
++        switch (state) {
++
++        case sw_nameserver:
++
++            if ((size_t) n - i >= sizeof("nameserver ") - 1
```
agentzh: Need to handle multiple spaces and/or tabs used before the word `nameserver`. I've tested and found that both Linux and Darwin do support such free formats.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -93,28 +101,53 @@ index cd55520c..6a3f98bb 100644
...
++            break;
++
 +        case sw_address:
 +
 +            if (buf[i] == CR || buf[i] == LF || i == n - 1) {
 +                ngx_memzero(&u, sizeof(ngx_url_t));
 +
 +                u.url.data = buf + address;
-+                u.url.len = i - address;
++                u.url.len = i == n - 1 ? n - address : i - address;
```
agentzh: Better add a parentheses around the condition to follow the nginx coding style.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -93,28 +101,53 @@ index cd55520c..6a3f98bb 100644
...
 +                u.default_port = 53;
 +
++                /* IPv6? */
++                if (ngx_strlchr(u.url.data, u.url.data + u.url.len, ':') != NULL) {
```
agentzh: This line exceeds 80 cols, it seems.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -93,28 +101,53 @@ index cd55520c..6a3f98bb 100644
...
++                    if (u.url.len + 2 > sizeof(ipv6_buf)) {
++                        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
++                                           "IPv6 resolver address is too long: \"%V\"",
```
agentzh: Ditto.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -164,15 +206,16 @@ index cd55520c..6a3f98bb 100644
...
+@@ -246,6 +429,35 @@ ngx_resolver_create(ngx_conf_t *cf, ngx_str_t *names, ngx_uint_t n)
          }
  #endif
  
 +        if (ngx_strncmp(names[i].data, "local=", 6) == 0) {
 +
 +            if (ngx_strcmp(&names[i].data[6], "on") == 0) {
 +                if (ngx_resolver_read_resolv_conf(cf, r,
-+                                                  (u_char *) NGX_RESOLVER_FILE_NAME)
++                                                  (u_char *) NGX_RESOLVER_FILE_NAME,
```
agentzh: Ditto.

patches/nginx-1.13.6-resolver_conf_parsing.patch
```diff
@@ -164,15 +206,16 @@ index cd55520c..6a3f98bb 100644
...
++                                                  sizeof(NGX_RESOLVER_FILE_NAME) - 1)
```
agentzh: Ditto.

agentzh:
@dndx Merged by renaming your test file to 001-xxx.t. Thanks!

PR #362, opened 2018-04-20
configure: put lua-resty libs into a list, and make all operations on lua-resty opts use that list

mclyne:
Reason for patch
=============
The configure file now has quite a large number of included lua-resty-[name] libs, with all references to them statically included in the following places:

1) Configure option parsing (each --without-lua_resty_xxx option)
2) Including the make files (a static list)
3) Configure usage message

This is both more prone to errors, and is inefficient when adding new libs in the future.

What this patch does
================
1) Puts all the lua-resty-[name] libs into a list (underneath the list of modules)
2) Uses the list to generate a regex to test for --without-lua_resty_[name] configure options
3) Uses the list to add the lua-resty-[name] libs to the whole $make process
4) Uses the list to generate the relevant parts of the ./configure --help message
5) A few very minor edits that improve clarity / formatting consistency (that make no changes to the actions)

This makes the whole configure script a bit easier to read, makes it quicker to add new lua-resty-[name] libs to OpenResty, and reduces the possibility for bugs in future edits.

What else you might want to change
============================
I've ordered the lua-resty-[name] modules alphabetically, because I think they're easier to read that way.  Given that the modules are not listed alphabetically in the ./configure --help, you might want to : 

- Re-order the list of lua-resty-[name] libs
- Modify the ./configure script to display the help with the --with-[name]_module and --without-[name]_module so they display alphabetically

I didn't want to presume your choice about this, but if you'd like me to do a quick patch to change that as well, I'm happy to do so.

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -88,6 +88,21 @@ my @modules = (
     [stream_lua => 'ngx_stream_lua'],
 );
 
+my @lua_resty_libs = (
+    'core',
+    'dns',
+    'limit_traffic',
+    'lock',
+    'lrucache',
+    'memcached',
+    'mysql',
+    'redis',
+    'string',
+    'upload',
+    'upstream_healthcheck',
+    'websocket',
+);
+
 my $without_resty_mods_regex;
 {
     my $s = '^--without-('
@@ -116,6 +131,17 @@ my $with_resty_mods_regex;
     #warn "with regex: $s";
 
     $with_resty_mods_regex = qr/$s/;
+}
+
+my $without_lua_resty_libs_regex;
+{
+    my $s = '^--without-lua_resty_('
+        . join('|',
+            @lua_resty_libs
+          )
+        . ')$';
+
+    $without_lua_resty_libs_regex = qr/$s/;
 }
 
 my $prefix = '/usr/local/openresty';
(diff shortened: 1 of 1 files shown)


mclyne:
Just to let you know, I double-checked to make sure that I hadn't forgotten any libs, and I tested with the following options :

``` 
    --without-lua_cjson
    --without-lua_redis_parser
    --without-lua_rds_parser
    --without-lua_resty_core
    --without-lua_resty_dns
    --without-lua_resty_limit_traffic
    --without-lua_resty_lock
    --without-lua_resty_lrucache
    --without-lua_resty_memcached
    --without-lua_resty_mysql
    --without-lua_resty_redis
    --without-lua_resty_string
    --without-lua_resty_upload
    --without-lua_resty_upstream_healthcheck
    --without-lua_resty_websocket
```
which then produced no lualib directory in $PREFIX folder.  Obviously leaving these options off installs all the libs, so functionally there is no change to behaviour, just a tidy-up of the script.

util/configure
```diff
@@ -370,7 +366,7 @@ push @make_install_cmds,
 
 if ($platform ne 'msys') {
     push @make_install_cmds,
-        "ln -sf $ngx_sbin \$(DESTDIR)$prefix/bin/openresty";
+        "ln -sf \$(DESTDIR)$ngx_sbin \$(DESTDIR)$prefix/bin/openresty";
```
agentzh: This is wrong. It will create a symlink pointing to the wrong location.

util/configure
```diff
@@ -1071,7 +1066,8 @@ _EOC_
         }
     }
 
-    # configure opm:
+    # configure opm
```
agentzh: Why this change?

util/configure
```diff
@@ -1084,7 +1080,7 @@ _EOC_
              . "$root_dir/build/install bin/* $target_dir";
     }
 
-    # configure resty-cli:
+    # configure resty-cli
```
agentzh: Ditto.

util/configure
mclyne: Consistency with other comments.  It was sort of an accidental inclusion.  Same for below.  Most comments don't have colons - though actually they're not totall consisten.

util/configure
mclyne: I guess it depends on where you want to have it link to, but if you have DESTDIR defined, then the symlink will link to the location without $(DESTDIR) added.  All other files are installed with $(DESTDIR) prefixed, so this probably isn't what you want.  Nginx is installed to the prefixed location.  Adding $(DESTDIR) to the $ngx_sbin path makes it point to the nginx binary in the same root folder in all situations.

There is also another problem if you have DESTDIR set.  That is that because the rpath and link locations are defined statically, nginx doesn't find libluajit, unless you specify it in the environment.  IMHO it would be better to have relative symlinks and relative paths set.  I was going to make the necessary changes to make them relative paths, so everything worked as expected if you move the output dir and submit them too.  It makes more sense to me to be able to move the whole folder and for everything still to work, but let me know if that's not what you want.

agentzh: @mclyne The `DESTDIR` variable is usually used for rpm/ubuntu packaging tools to install the software to a temporary location other than under the real root directory. So it is not meant to be the final location for the installed packages. That's why I said if you add `$(DESTDIR)` to the symlink target location, it would be completely wrong (the DESTDIR location won't even exist in the final deployment machine).

agentzh: @mclyne Seems like you confused DESTDIR with PREFIX. They are for completely different purposes.

agentzh: @mclyne So your change above would lead to broken symlinks installed to the final machines after installed the deb/rpm packages.

mclyne: I wasn't confusing DESTDIR with PREFIX, however I wasn't thinking of it being used in that way.  Thanks for the explanation.

PR #396, opened 2018-08-07
feature: added the preserve_method_on_error_page patch.

thibaultcha:
> I hereby granted the copyright of the changes in this pull request

Wondering if this is something that we may want; please advise.

In cases where we want to access to a yield-able Lua context upon Nginx-produced or upstream errors, the `error_page` directive is our de-facto choice. However, when logging such errors (along with the request that triggered it) to an external logging stack, the request's HTTP method is overridden by Nginx to `GET`, causing us to lose some context about the original request.
My guess is that the override makes sense by default since most Nginx instances would then serve their HTML error page upon this internal redirect. I tracked down the behaviour from a release dating back to 2005:

https://github.com/nginx/nginx/commit/ceb992921cee6f76d1752af2d388ee6a1d71e078#diff-05d6b8ae21d192f78b349319edaeb9bfR295

However, there seems to be no particular reason why this would have to be the case when one produces their own content in the `error_page` location, e.g. via `content_by_lua_*`. Hence, a directive with backwards-compatibility defaults seems to make sense.

### DIFF
--- a/patches/nginx-1.13.6-preserve_method_on_error_page.patch
+++ b/patches/nginx-1.13.6-preserve_method_on_error_page.patch
@@ -1,0 +1,93 @@
+# HG changeset patch
+# User Thibault Charbonnier <thibaultcha@me.com>
+# Date 1532789509 -7200
+#      Sat Jul 28 16:51:49 2018 +0200
+# Node ID af4366d9fec2bc2ccce8fd03fc456621df13f673
+# Parent  f7e79596baf209151682f2f7d220161c034657ac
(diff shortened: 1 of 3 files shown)


jeremyjpj0916:
+1 , any updates around this? We still get week to week confusion from users seeing HTTP GET method in the logs with the reality was the client sent a POST/DELETE/PUT etc. and a L4 connection failure occurred so it was mis-logged as a GET.

thibaultcha:
Closing this as we will be addressing the issue differently in Kong.

PR #447, opened 2019-01-25
change: we now enable GC64 mode by default in our bundled LuaJIT build for x86_64 architectures [...]

thibaultcha:
[...] this can be disabled with ./configure --without-luajit-gc64.

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -4,6 +4,7 @@ use 5.006;
 use strict;
 use warnings;
 
+use Config;
 use File::Spec;
 use File::Temp qw( tempfile tmpnam );
 
@@ -129,6 +130,7 @@ my $cores;
 my $luajit_xcflags = '';
 my $user_luajit_xcflags;
 my $no_luajit_lua52;
+my $no_luajit_gc64;
 
 my (@ngx_opts, @ngx_cc_opts, @ngx_ld_opts);
 
@@ -262,6 +264,9 @@ for my $opt (@ARGV) {
 
     } elsif ($opt =~ /^--without-luajit-lua52/) {
         $no_luajit_lua52 = 1;
+
+    } elsif ($opt =~ /^--without-luajit-gc64/) {
+        $no_luajit_gc64 = 1;
 
     } elsif ($opt =~ /^--with-libdrizzle=(.*)/) {
         $resty_opts{libdrizzle} = $1;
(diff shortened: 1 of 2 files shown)


util/configure
```diff
@@ -697,6 +701,14 @@ _END_
             $luajit_xcflags .= " -DLUAJIT_ENABLE_LUA52COMPAT";
         }
 
+        if (!$no_luajit_gc64
+            && `uname -a` =~ /\bx86_64\b/
```
thibaultcha: Note: iirc, the GC64 mode is required on ARM64.

agentzh: I think we should use Perl's own architecture information instead of invoking `uname` which is not available everywhere.

BTW, GC64 is required on ARM64 so we do not need to worry about that at all. The LuaJIT build system enables that automatically anyway.

util/configure
```diff
@@ -1302,6 +1314,7 @@ _EOC_
   --with-luajit-xcflags=FLAGS        Specify extra C compiler flags for LuaJIT 2.1
   --without-luajit-lua52             Turns off the LuaJIT extensions from Lua 5.2 that may break
                                      backward compatibility.
+  --without-luajit-gc64              Turns off LuaJIT GC64 mode
```
agentzh: We should mention x86_64 here?

util/configure
thibaultcha: Should we then update existing invocations of `uname`? i.e. https://github.com/openresty/openresty/blob/feb1bc4c2a060abf48c24a5cdbe685f156e80829/util/configure#L325

agentzh: We could though it's less an issue since it first makes sure it is on darwin.

thibaultcha: Updated

util/configure
thibaultcha: Good point, updated

util/configure
thibaultcha: > The LuaJIT build system enables that automatically anyway.

That's helpful, thanks, I was about to get my hands on an ARM64 box to test this (still planning on doing so today or tomorrow).

agentzh: @thibaultcha Great. Thanks! It would be awesome if you could run the tests on 32-bit arm systems at the same time, to cover corner cases as in openresty/luajit2#37.

PR #473, opened 2019-04-10
bugfix: we don't download tar.bz2 archive anymore.

spacewander:
Therefore, there is no need to run the 'rm' command.

### DIFF
--- a/util/mirror-tarballs
+++ b/util/mirror-tarballs
@@ -884,7 +884,6 @@ cd ..
 #################################
 
 rm *.tar.gz
-rm *.tar.bz2
 
 cd ..
 cp $root/util/configure ./ || exit 1


spacewander:
The CI failure is relative with https://github.com/openresty/resty-cli/pull/49

thibaultcha:
This was just added a few months ago: https://github.com/openresty/openresty/commit/c441759ce30298b965830ff3bd8a8f3a7e3eb04b. What's more interesting, is that at the time of addition, we were already not explicitly downloading any tar.bz2 archives. Therefore, this raises the question: did we miss something here and this command may be needed in some cases?

agentzh:
@thibaultcha I'd just keep it for the future.

PR #474, opened 2019-04-10
feature: supported async SSL session fetch callback with OpenSSL 1.1.1.

spacewander:


spacewander:
Maybe we could rename the `ssl_pending_session.patch` to `ssl_sess_cb_yield.patch`? It is not just about the `pending session` feature.

agentzh:
@spacewander Yeah, we could rename the patch file.

spacewander:
@agentzh 
OK, I will do it after this patch is merged.

patches/nginx-1.15.8-ssl_pending_session.patch
```diff
@@ -1,16 +1,36 @@
...
+diff --git a/src/event/ngx_event_openssl.h b/src/event/ngx_event_openssl.h
+--- a/src/event/ngx_event_openssl.h
++++ b/src/event/ngx_event_openssl.h
+@@ -59,6 +59,11 @@
+ #endif
+ 
+ 
++#ifdef SSL_ERROR_WANT_CLIENT_HELLO_CB
++#define HAVE_SSL_CLIENT_HELLO_CB_SUPPORT  1
++#endif
```
thibaultcha: Is this alias truly necessary? It increases conflict surface with NGINX and adds indirection which makes it harder for the reader to understand.
[Here](https://github.com/openresty/lua-nginx-module/pull/1503/files#diff-c91b2ec425e9be9d56fee2df662b67d2R1081), the patch uses `SSL_ERROR_WANT_CLIENT_HELLO_CB` directly. Also, [here](https://github.com/openresty/lua-nginx-module/pull/1503/files#diff-627bc846d080debcf2f52ff0ae26e6e9R343) we use another `SSL_ERROR_*` macro already.

patches/nginx-1.15.8-ssl_pending_session.patch
```diff
@@ -1,16 +1,36 @@
...
      }
  
  #if OPENSSL_VERSION_NUMBER >= 0x10002000L
 -    if (sslerr == SSL_ERROR_WANT_X509_LOOKUP) {
 +    if (sslerr == SSL_ERROR_WANT_X509_LOOKUP
 +#   ifdef SSL_ERROR_PENDING_SESSION
 +        || sslerr == SSL_ERROR_PENDING_SESSION
++
++#   else
++#       ifdef SSL_ERROR_WANT_CLIENT_HELLO_CB
```
thibaultcha: Instead of:
```c
#   ifdef SSL_ERROR_PENDING_SESSION
          // ...

#   else
#       ifdef SSL_ERROR_WANT_CLIENT_HELLO_CB
         // ...
#       endif
#   endif
```

How about:
```c
#   ifdef SSL_ERROR_PENDING_SESSION
        // ...

#   elif defined(SSL_ERROR_WANT_CLIENT_HELLO_CB)
        // ...
#   endif
```

A pattern that NGINX already uses.

patches/nginx-1.15.8-ssl_pending_session.patch
```diff
@@ -1,16 +1,36 @@
 --- nginx-1.15.8/src/event/ngx_event_openssl.c	2016-07-17 19:20:30.411137606 -0700
 +++ nginx-1.15.8-patched/src/event/ngx_event_openssl.c	2016-07-19 16:53:35.539768477 -0700
-@@ -1307,7 +1307,12 @@ ngx_ssl_handshake(ngx_connection_t *c)
+@@ -1307,7 +1307,17 @@ ngx_ssl_handshake(ngx_connection_t *c)
```
thibaultcha: minor: On NGINX's `release-1.15.8` tag with the `nginx-1.15.8-ssl_cert_cb_yield` patch already applied, I get different line numbers for this chunk once I apply this patch:
```
@@ -1446,7 +1446,17 @@ ngx_ssl_handshake(ngx_connection_t *c)
```

This doesn't seem to prevent the patch from applying, but worth mentioning.

patches/nginx-1.15.8-ssl_pending_session.patch
spacewander: @thibaultcha 
`HAVE_SSL_CLIENT_HELLO_CB_SUPPORT` indicates two things:
1. `SSL_ERROR_WANT_CLIENT_HELLO_CB` is defined.
2. This patch is applied.

Using `SSL_ERROR_WANT_CLIENT_HELLO_CB` barely is incorrect, I will fix the lua-nginx-module's PR later.

spacewander:
I have submitted another PR to fix the CI: https://github.com/openresty/openresty/pull/492

patches/nginx-1.15.8-ssl_pending_session.patch
thibaultcha: > 2. This patch is applied.

Good point, fair enough!

agentzh:
@spacewander Please update the `util/mirror-tarballs` script (as well as the `ngx-build` script in openresty-devel-utils) to actually apply this patch to the nginx core.

spacewander:
@agentzh @thibaultcha 
I will rename this `ssl_pending_session.patch` to `ssl_sess_cb_yield.patch` as previous discussed, but in this PR.

spacewander:
I have submitted a PR to `openresty-devel-utils`: https://github.com/openresty/openresty-devel-utils/pull/22, and also updated this one.

thibaultcha:
Merged!

PR #478, opened 2019-04-20
feature: closed unused listening fds in worker processes.

spacewander:
When 'reuseport' is enabled in the 'listen' directive, Nginx
will create a listening fd for each worker process in the
master process.

These fds will be inherited by the worker processes, but
most of them are unused. For eaxmple, considered we have
32 listening ip:port configurations and 64 worker processes,
each worker process will inherit 2048 (32 * 64) listening fds,
but only 32 fds are used. By closing the unused fds, this
change could save up to 2016 (32 * 63) fds in a worker process.

It doesn't affect the listening socket, since there is
only one used fd which associates to the socket with or
without this change.


thibaultcha:
Manually merged with minor edits and new tests. Thanks!

PR #497, opened 2019-06-15
bugfix: supported OpenSSL 1.1.1 by upgrading the OpenSSL patch.

spacewander:
Previously, we used ClientHello callback to do ssl session fetching
non-blockingly. However, this way can not handle an edge case: the ssl
session resumption via session ticket might fail, and the client falls
back to session ID resumption. The ClientHello callback is run too early
to know if the client will fall back to use session ID resumption.

Therefore, we have to take back the OpenSSL sess_set_get_cb_yield patch and
upgrade it to adapt OpenSSL 1.1.1.

Thanks Yongjian Xu and crasyangel for their help.

### DIFF
--- a/patches/nginx-1.15.8-ssl_sess_cb_yield.patch
+++ b/patches/nginx-1.15.8-ssl_sess_cb_yield.patch
@@ -1,6 +1,7 @@
---- nginx-1.15.8/src/event/ngx_event_openssl.c	2016-07-17 19:20:30.411137606 -0700
-+++ nginx-1.15.8-patched/src/event/ngx_event_openssl.c	2016-07-19 16:53:35.539768477 -0700
-@@ -1581,7 +1581,15 @@ ngx_ssl_try_early_data(ngx_connection_t *c)
+diff --git a/src/event/ngx_event_openssl.c b/src/event/ngx_event_openssl.c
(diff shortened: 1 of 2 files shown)


thibaultcha:
LGTM. One comment however: since we are not planning on releasing another 1.15.8 release, I would like to revert the patches to what they were when we released 1.15.8, and add such support for OpenSSL 1.1.1 starting with OpenResty 1.17.x.1. See the PR for 1.17.1 support: https://github.com/openresty/openresty/pull/503.

This ensures that the state of the 1.15.8 patches remain the same, and prevents confusion as to which patches shipped with 1.15.8.1 (currently, the patches don't reflect the contents of 1.15.8.1 release artefacts, and unless we plan on 1.15.8.2, which we don't, they should reflect what shipped in 1.15.8.1).

spacewander:
@thibaultcha 
Maybe we should get https://github.com/openresty/openresty/pull/503 merged first, so that I can rebase this PR.

BTW, we need to address https://github.com/openresty/openresty/pull/499 before making a new release.

thibaultcha:
@spacewander Yes, both are on my list

thibaultcha:
Merged, thank you!

Issue #540, opened 2019-10-03
https://openresty.org/download/nginx-1.17.4.tar.gz is missing


arcivanov:
The tree as of 7508c18 is missing nginx-1.17.4.tar.gz tarball referenced in  https://github.com/openresty/openresty/blob/master/util/mirror-tarballs#L31

mhf-ir:
How much longer nginx 1.17 to be ready for https://github.com/openresty/openresty-packaging ?
Any schedule for that ?

TheNorthMemory:
the origin nginx source codes had been some changes since 1.17.1 to 1.17.4, this commit e997370 is solved the hang of 7508c18 like below:

```
[INFO] applying the stream_ssl_preread_no_skip patch for nginx
patching file src/stream/ngx_stream_ssl_preread_module.c
Hunk 1 FAILED 159/159.

         rc = ngx_stream_ssl_preread_parse_record(ctx, p, p + len);
         if (rc != NGX_AGAIN) {
-            return rc;
+            return rc == NGX_OK ? NGX_DECLINED : rc;
         }

         p += len;
make: *** [Makefile:4: all] Error 1
```

handy packaging steps as:

```
git clone --single-branch --depth 1 --branch release-1.17.4 https://github.com/nginx/nginx.git nginx/nginx-1.17.4
cp -af nginx/nginx-1.17.4/docs/text/README nginx/nginx-1.17.4/
cp -af nginx/nginx-1.17.4/auto/configure nginx/nginx-1.17.4/
tar --exclude '.git' -C nginx -czf nginx-1.17.4.tar.gz nginx-1.17.4
git clone --depth=50 https://github.com/openresty/openresty.git openresty/openresty
cd openresty/openresty
git fetch origin +refs/pull/558/merge:
git checkout FETCH_HEAD
mkdir -p ./work
mv ../../nginx-1.17.4.tar.gz ./work
make
```

after the `make` done, the package should be there as `openresty-1.17.4*.tar.gz`

wants to switch the minor version (current `1rc0`@ 7508c18 ) of the openresty, just `sed` the def `minor_ver` in the file `util/ver` as you want.

feel free to try it.

agentzh:
One should use this tarball:

https://openresty.org/download/nginx-1.17.4.tar.gz

It was uploaded for several weeks already. Consider it resolved.

PR #558, opened 2019-12-02
fix openresty/openresty#540 manually

TheNorthMemory:
tested under alpine as:

```
FROM alpine
COPY . /openresty
RUN apk add --no-cache bash git make gzip perl mercurial
  && cd /openresty
  && make
  && ls *.tar.gz|xargs -rI {} tar -ztvf {}
```


TheNorthMemory:
this did more changes, there should be a smart pr soon.

PR #560, opened 2019-12-07
repair the patches of 1.17.4 rc0, let things work

TheNorthMemory:

A minimal repairing refer to #558, checked those commits
- nginx/nginx@8d97a2e
- nginx/nginx@c60b61a

Also a little bit fixed on `TEST 22` on the alpine linux, all test cases both passed under the alpine and ubuntu.

Some notes were in each commits below.


patches/nginx-1.17.4-dtrace.patch
```diff
@@ -1074,16 +1074,8 @@ index de1b2027..887cf6cd 100644
...
-     }
- }
- 
--
- ngx_int_t
- ngx_http_process_request_uri(ngx_http_request_t *r)
- {
 @@ -1348,6 +1348,8 @@ ngx_http_process_request_headers(ngx_event_t *rev)
-                 return;
+                 break;
```
agentzh: Will you please explain these changes a bit?

patches/nginx-1.17.4-gcc-maybe-uninitialized-warning.patch
```diff
@@ -7,5 +7,5 @@
 -    ngx_http_core_srv_conf_t  *cscf;
 +    ngx_http_core_srv_conf_t  *cscf = NULL;
  
-     hc = r->http_connection;
- 
+ #if (NGX_SUPPRESS_WARN)
+     cscf = NULL;
```
agentzh: Is it because of new changes in the nginx core?

patches/nginx-1.17.4-intercept_error_log.patch
```diff
@@ -3,7 +3,7 @@ index c51b7ff..4c335b9 100644
 --- a/src/core/ngx_cycle.h
 +++ b/src/core/ngx_cycle.h
 @@ -22,9 +22,14 @@
- #define NGX_DEBUG_POINTS_ABORT  2
+ #define HAVE_PRIVILEGED_PROCESS_PATCH   1
```
agentzh: What is this change?

patches/nginx-1.17.4-gcc-maybe-uninitialized-warning.patch
TheNorthMemory: This was changed since 2014, please check nginx/nginx@8d97a2e. In my opition, this patch should be EOL(the mention was also committed as the message body).

patches/nginx-1.17.4-dtrace.patch
TheNorthMemory: This one was changed here nginx/nginx@e4a3211e, fixed the socket leak with "return 444" in error_page.

patches/nginx-1.17.4-intercept_error_log.patch
TheNorthMemory: The `src/core/ngx_cycle.h` was patched by the `privileged-agent-process` before

https://github.com/openresty/openresty/blob/7508c1852265bd04fdb2dfd64989d4c490440f1a/patches/nginx-1.17.4-privileged_agent_process.patch#L25-L34

TheNorthMemory:
EOL

Issue #587, opened 2020-02-20
Cannot add dynamic module into openresty - nginx


satscreate:
**I installed a latest version:**
`https://openresty.org/download/openresty-1.15.8.2.tar.gz`

**nginx -V**
```
nginx version: openresty/1.15.8.2
built by gcc 4.8.5 20150623 (Red Hat 4.8.5-39) (GCC)
built with OpenSSL 1.0.2k-fips  26 Jan 2017
TLS SNI support enabled
configure arguments: --prefix=/usr/local/openresty/nginx --with-cc-opt=-O2 --add-module=../ngx_devel_kit-0.3.1rc1 --add-module=../echo-nginx-module-0.61 --add-module=../xss-nginx-module-0.06 --add-module=../ngx_coolkit-0.2 --add-module=../set-misc-nginx-module-0.32 --add-module=../form-input-nginx-module-0.12 --add-module=../encrypted-session-nginx-module-0.08 --add-module=../srcache-nginx-module-0.31 --add-module=../ngx_lua-0.10.15 --add-module=../ngx_lua_upstream-0.07 --add-module=../headers-more-nginx-module-0.33 --add-module=../array-var-nginx-module-0.05 --add-module=../memc-nginx-module-0.19 --add-module=../redis2-nginx-module-0.15 --add-module=../redis-nginx-module-0.3.7 --add-module=../rds-json-nginx-module-0.15 --add-module=../rds-csv-nginx-module-0.09 --add-module=../ngx_stream_lua-0.0.7 --with-ld-opt=-Wl,-rpath,/usr/local/openresty/luajit/lib --with-stream --with-stream_ssl_module --with-stream_ssl_preread_module --with-http_ssl_module
```

Happily nginx starts with below command,
`sudo systemctl restart nginx.service`

now  when i try to add some modules, like https://nginx.org/en/docs/njs/install.html#install_sources

i tried,
```
 [openresty-1.15.8.2]# ./configure --with-compat --add-dynamic-module=../njs-0173143d7b15/nginx
.....................
make & make install
```

Then,
make results in below error, adding few lines.. 

```
gmake[1]: Entering directory `/tmp/openresty/openresty-1.15.8.2/build/nginx-1.15.8'
gmake -f objs/Makefile
gmake[2]: Entering directory `/tmp/openresty/openresty-1.15.8.2/build/nginx-1.15.8'
cd /tmp/openresty/openresty-1.15.8.2/../njs-0173143d7b15/nginx/.. \
[...]

PR #591, opened 2020-02-28
bugfix: use to LUA_ORIGIN_DEFAULT_CPATH instead of LUA_DEFAULT_PATH avoid dquote in CFLAGS

rainingmaster:
Fix this issue: https://github.com/openresty/openresty/issues/587.

The reason for this problem is that escape will work in shell, but not work in the Makefile, such as [this example](https://gist.github.com/doujiang24/fe0c541a585a300fcc3265a4dede0837). When Nginx's makefile is generated, the config.make file of the third-party module is called [here](https://github.com/nginx/nginx/blob/master/auto/make#L445), and the third-party module may use CFLAGS in [config.make](https://github.com/nginx/njs/blob/master/nginx/config.make#L6). When dquote exists in CFLAGS, it will cause an error when executing make, such as this [Makefile](https://gist.github.com/rainingmaster/f5421344a65de13a380d229c261cd537#file-makefile-L3), the CFLAGS is `... -DLUA_DEFAULT_PATH='"/usr/local/openresty/site/lualib/?.ljbc;...`, so build will be broken because CFLAGS will be split to some string in [build/Makefile](https://gist.github.com/rainingmaster/f5421344a65de13a380d229c261cd537#file-makefile-L3007)

Thanks for @doujiang24 's help, we found some way to fix this problem:
1. [By this way](https://github.com/openresty/openresty/issues/587#issuecomment-589814604), we try to add some patch to original NGINX, but found the wrong make file is generate from third part module, such as njs, it is not easy to patch.Another weird way is add some code to escape the CFLAGS when using  build/Makefile in the nginx/objs/Makefile at [nginx's configure](https://github.com/nginx/nginx/blob/master/auto/configure#L105). This solution has not been implemented for the time being.
1. Change the way of define `LUA_DEFAULT_PATH`, such as [this way](https://github.com/rainingmaster/openresty/commit/844b60fd0c0872908cb579f066bf87aee0b3a4ed#diff-7b0d055ed948accb24a95b6de87685b5R965). However, `#` is also need to escape, so when from Makefile(nginx) -> config.make(njs) -> Makefile(njs), it will not work again.
1. It is this PR, we change the way to define the `LUA_DEFAULT_PATH`, such as [this](https://github.com/rainingmaster/lua-nginx-module/commit/9825d0bfa625048f9cff5b062b304074e5af475f#diff-4c237c57e5462e077417a5177672ce82R157), but it will be very complicated because `lua-nginx-module` and `stream-lua-nginx-module` also need to be changed.

@agentzh Do you have any other idea to solve this? Thanks if you can share.

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -962,9 +962,8 @@ int main(void) {
         {
             print $in <<"_EOC_";
 
-ngx_lua_dquote='"'
-CFLAGS="\$CFLAGS -DLUA_DEFAULT_PATH='\${ngx_lua_dquote}$site_lualib_prefix/?.ljbc;$site_lualib_prefix/?/init.ljbc;$lualib_prefix/?.ljbc;$lualib_prefix/?/init.ljbc;$site_lualib_prefix/?.lua;$site_lualib_prefix/?/init.lua;$lualib_prefix/?.lua;$lualib_prefix/?/init.lua\$ngx_lua_dquote'"
-CFLAGS="\$CFLAGS -DLUA_DEFAULT_CPATH='\${ngx_lua_dquote}$site_lualib_prefix/?.so;$lualib_prefix/?.so\$ngx_lua_dquote'"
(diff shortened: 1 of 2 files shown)


agentzh:
@rainingmaster Sorry for the late reply.

After some thinking, I'd rather automatically generate a header file with those macro definitions to avoid `'` or `"` altogether in the `CFLAGS` value. I've got a patch working locally. I'll commit it tomorrow.

agentzh:
We now merged a patch taking a different (and better) approach.

PR #600, opened 2020-03-31
Update bundled openresty/luajit2 to v2.1-20200102

mattarnoatibm:
Update the version of openresty/luajit2 in the 1.15.8.x branch to
v2.1-20200102.

This is to allow a release of openresty that includes a version of
luajit that supports the s390x architecture to be created. Support for
s390x was added to openresty/luajit2 by this PR -

https://github.com/openresty/luajit2/pull/50

which is included in openresty/luajit2 release v2.1-20200102, however
the bundled release of openresty/luajit2 in the latest release of
openresty, version 1.15.8.3, is v2.1-20190507, which has no support
for s390x.

Commit added by Matthew Arnold <mattarno@uk.ibm.com>

### DIFF
--- a/t/sanity.t
+++ b/t/sanity.t
@@ -276,7 +276,7 @@ platform: linux (linux)
 platform: linux (linux)
 cp -rp bundle/ build
 cd build
-cd LuaJIT-2.1-20190507
+cd LuaJIT-2.1-20200102
 INFO: found -msse4.2 in cc.
 gmake TARGET_STRIP=@: CCDEBUG=-g XCFLAGS='-DLUAJIT_ENABLE_LUA52COMPAT -DLUAJIT_ENABLE_GC64 -msse4.2' CC=cc PREFIX=/usr/local/openresty/luajit
 gmake install TARGET_STRIP=@: CCDEBUG=-g XCFLAGS='-DLUAJIT_ENABLE_LUA52COMPAT -DLUAJIT_ENABLE_GC64 -msse4.2' CC=cc PREFIX=/usr/local/openresty/luajit DESTDIR=$OPENRESTY_BUILD_DIR/luajit-root/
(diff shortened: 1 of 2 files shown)


thibaultcha:
Hi,

While not personally opposed to the bump (although I do have some concerns since our extensive EC2-hosted test suite typically gets ran before such releases bumping the LuaJIT version), what prevents you from building OpenResty with LuaJIT v2.1-20200102? You can compile OpenResty and use the `with-luajit=DIR` option for the time being, regardless of whether this bump makes it to 1.15.8.x or not.

mattarnoatibm:
We are working on adding s390x support to ingress-nginx controller, https://github.com/kubernetes/ingress-nginx, which uses openresty. This requires an official openresty build, rather than a build I have built myself outside the official builds.

How do we run the EC2-hosted test suite? Could we run it against a build of openresty from 1.15.8.x, with LuaJIT v2.1-20200102, and see if we can bump the version of LuaJIT based on that?

thibaultcha:
> This requires an official openresty build, rather than a build I have built myself outside the official builds.

How come? Why not downloading an official build, and compiling it yourself against the version of LuaJIT you want to use? You'd be unblocked without having to wait for a new release.

mattarnoatibm:
We want add s390x support to the Nginx Ingress Controller for Kubernetes project. Not just build it for s390x for myself. To make that change in the repo I linked above it's going to be necessary to have an official openresty release that includes a version of LuaJIT that supports s390x.

Could I understand what the planned schedule for publishing the next release of openresty is?

At the moment, the main issue for me I'm concerned about is not the wait for the next official release of openresty to get published, but that the next official release of openresty will still come bundled with LuaJIT version 2.1-20190507, which doesn't support s390x. Thus the next official release would still not be usable for building Nginx Ingress Controller for s390x.

aledbf:
@mattarnoatibm two comments:

> We are working on adding s390x support to ingress-nginx controller, https://github.com/kubernetes/ingress-nginx, which uses openresty.

Since [0.27.1](https://github.com/kubernetes/ingress-nginx/blob/master/Changelog.md#0270), ingress-nginx does not use openresty anymore. We do use LuaJIT and lua-nginx-module.

> We want add s390x support to the Nginx Ingress Controller for Kubernetes project.

Unless you can provide long term access to s390x hardware, we cannot add support for new platforms.

Also, to complement what @thibaultcha said, you can fork ingress-nginx and build a custom nginx image, just changing [this line](https://github.com/kubernetes/ingress-nginx/blob/master/images/nginx/rootfs/build.sh#L47) pointing to the correct commit. Once you have done that, you need to point the ingress controller to use that image [here](https://github.com/kubernetes/ingress-nginx/blob/master/Makefile#L66-L67)

edmondsw:
I did a little digging and found the change Manuel was talking about where ingress-nginx switched from openresty to nginx, but does still use openresty/luajit2: https://github.com/kubernetes/ingress-nginx/pull/4863

@mattarnoatibm hope that helps.

mattarnoatibm:
I'll see if we can get a build based on the `0.30.0` release of ingress-nginx to build for s390x. If that works we can proceed without openresty.

thibaultcha:
For the record, the next major OpenResty release will be 1.17.8.1, of which [RC1](http://openresty.org/en/ann-1017008001rc1.html) is currently available right now. It comes bundled with LuaJIT 2.1-20200102 which has s390x support. But again: there was no need for an official OpenResty release to bundle it in the first place since you can simply compile it yourself, quite easily, with a single `configure` option.

Thanks for the info @aledbf!

abhay1722:
> For the record, the next major OpenResty release will be 1.17.8.1, of which [RC1](http://openresty.org/en/ann-1017008001rc1.html) is currently available right now. It comes bundled with LuaJIT 2.1-20200102 which has s390x support. But again: there was no need for an official OpenResty release to bundle it in the first place since you can simply compile it yourself, quite easily, with a single `configure` option.
> 
> Thanks for the info @aledbf!

@thibaultcha Just for confirmation, Does this mean the next major OpenResty release i.e. 1.17.8.1 provide official [pre-built packages](https://openresty.org/en/linux-packages.html) for s390x as well?

abhay1722:
> Unless you can provide long term access to s390x hardware, we cannot add support for new platforms.

@aledbf We can provide s390x hardware. Please let us know.

aledbf:
> We can provide s390x hardware. Please let us know.

@abhay1722 please send me an email aledbf@gmail.com

PR #612, opened 2020-05-15
add -g option for openssl pcre zlib

zhuizhuhaomeng:

auto add  -g  option no matter  user config with or without --with-(openssl|zlib|pcre)-opt

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -354,7 +354,7 @@ for my $opt (@ARGV) {
     } elsif ($opt =~ /^--with-(openssl|pcre|zlib)-opt=(.*)/) {
         my ($lib, $opt) = ($1, $2);
         push @ngx_opts, "--with-$lib-opt=-g $opt";
-        $with_ext_lib_opts{"$lib-opt"} = 1;
+        $with_ext_lib_opts{$lib} = 1;
 
     } elsif ($opt =~ /^--sbin-path=(.*)/) {
         $ngx_sbin = $1;
@@ -369,7 +369,7 @@ for my $opt (@ARGV) {
 }
 
 for my $lib (qw/openssl pcre zlib/) {
-    if ($with_ext_lib{$lib} && !$with_ext_lib_opts{"$lib-opt"}) {
+    if ($with_ext_lib{$lib} && !$with_ext_lib_opts{$lib}) {
         push @ngx_opts, "--with-$lib-opt=-g";
     }
 }
--- a/util/configure
+++ b/util/configure
@@ -25,6 +25,9 @@ my $OS = $^O;
 
 my ($ngx_dir, $lualib_prefix);
 
+my %with_ext_lib;
+my %with_ext_lib_opts;
+
 for my $opt (@ARGV) {
     if ($opt =~ /^--platform=(.*)/) {
         $OS = $1;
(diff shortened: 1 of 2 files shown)


util/configure
```diff
@@ -359,6 +371,15 @@ for my $opt (@ARGV) {
     }
 }
 
+
+if ($OS eq "linux") {
+        for my $lib (qw(openssl pcre zlib)) {
```
doujiang24, line 376: style: bad indent.

util/configure
```diff
@@ -359,6 +369,12 @@ for my $opt (@ARGV) {
     }
 }
 
+for my $lib (qw(openssl pcre zlib)) {
```
agentzh, line 372: Better use `qw/.../` instead of `qw(...)` for better readability here.

util/configure
```diff
@@ -359,6 +369,12 @@ for my $opt (@ARGV) {
...
+    if (exists($ngx_libs{$lib}) && not exists($ngx_libs_opt{$lib."-opt"})) {
```
agentzh, line 373: Please use `!` instead of `not` for expression level operators.

util/configure
```diff
@@ -359,6 +369,12 @@ for my $opt (@ARGV) {
...
+        push @ngx_opts, "--with-".$lib."-opt=-g";
```
agentzh, line 374: Please always use spaces around binary operators like `.`.

util/configure
```diff
@@ -346,6 +349,13 @@ for my $opt (@ARGV) {
         }
         $path = File::Spec->rel2abs($path);
         push @ngx_opts, "--with-$lib=$path";
+        $ngx_libs{$lib} = "y";
+
+    } elsif ($opt =~ /^--with-(openssl|pcre|zlib)-opt=(.*)/) {
+        my ($lib, $opt) = ($1, $2);
+        $opt.=" -g";
```
agentzh, line 356: Missing spaces around binary operator `.=`. Please fix other similar issues in your patch here. I see several such things below as well.

util/configure
agentzh: The logic here does not look quite right. Even if the user specifies her own `--with-xxx-opt`, we should still prepend `-g` to the option string.

util/configure
```diff
@@ -346,6 +349,13 @@ for my $opt (@ARGV) {
...
+        push @ngx_opts, "--with-$lib-opt=$opt";
+        $ngx_libs_opt{$lib."-opt"} = "y";
```
agentzh, line 358: We are not using the value `y` in any way, so a Perl boolean value `1` should be enough here?

util/configure
```diff
@@ -346,6 +349,12 @@ for my $opt (@ARGV) {
         }
         $path = File::Spec->rel2abs($path);
         push @ngx_opts, "--with-$lib=$path";
+        $with_ext_lib{$lib} = 1;
+
+    } elsif ($opt =~ /^--with-(openssl|pcre|zlib)-opt=(.*)/) {
+        my ($lib, $opt) = ($1, $2);
+        push @ngx_opts, "--with-$lib-opt=-g $opt";
+        $with_ext_lib_opts{"$lib-opt"} = 1;
```
agentzh, line 357: We don't need to add the `-opt` suffix to the hash table key. When we look up this table, we use the key of `$lib` anyway. It's wasteful to always append this artificial `-opt` suffix for no gains.

util/configure
```diff
@@ -359,6 +368,12 @@ for my $opt (@ARGV) {
     }
 }
 
+for my $lib (qw/openssl pcre zlib/) {
+    if ($with_ext_lib{$lib} && !$with_ext_lib_opts{"$lib-opt"}) {
```
agentzh, line 372: Ditto, let's save the trouble of constructing the key `"$lib-opt"`. Just use `$lib` instead.

agentzh:
Merged. Thanks.

PR #642, opened 2020-07-30
feature: --ngx_prefix for specific prefix for nginx

rainingmaster:
Sometime we can to use the binary of Openresty instead of the binary of Nginx, but we have many old config which the path is related to Nginx's prefix. So we need specific the prefix of Nginx, but not `$prefix_of_openresty/nginx`.

### DIFF
--- a/util/configure
+++ b/util/configure
@@ -124,6 +124,7 @@ my $with_resty_mods_regex;
 }
 
 my $prefix = '/usr/local/openresty';
+my $ngx_prefix;
 my $ngx_sbin;
 my %resty_opts;
 my $dry_run;
@@ -171,6 +172,9 @@ for my $opt (@ARGV) {
         if ($prefix eq '') {
             $prefix = '.';
         }
+
+    } elsif ($opt =~ /^--ngx-prefix=(.*)/) {
+        $ngx_prefix = $1;
 
     } elsif ($opt eq '--without-lua51') {
         die "ERROR: --without-lua51 is no longer supported.\n";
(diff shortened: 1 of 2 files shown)


util/configure
```diff
@@ -172,6 +173,9 @@ for my $opt (@ARGV) {
             $prefix = '.';
         }
 
+    } elsif ($opt =~ /^--ngx_prefix=(.*)/) {
```
xiaocang, line 176: `--ngx-prefix` should be used instead of `--ngx_prefix` in the command line options.

util/configure
```diff
@@ -381,10 +385,9 @@ if ($platform eq 'msys') {
 
 print "platform: $platform ($OS)\n";
 
-my $ngx_prefix;
 if ($platform eq 'msys') {
     $ngx_prefix = "$prefix";
-} else {
+} elsif (!$ngx_prefix) {
```
xiaocang, line 390: It' s weird: can't `msys` platform specify `$nginx_prefix` using the command line option?

xiaocang (changes requested):

util/configure
rainingmaster: updated

xiaocang (approved):

agentzh:
I don't like this. This will break the `resty` tool.

rainingmaster:
> I don't like this. This will break the `resty` tool.

@agentzh Hi, as my understand, this is a new feature, and will not affect any old tools and logic. Could you help explain what break `resty` mentioned above refers to? Thanks

PR #1083, opened 2025-11-04
bugfix: wake parent request when HTTP/2 subrequest finishes with error

tzssangglass:
When a client disconnects during an HTTP/2 subrequest (e.g., while ngx.sleep() is yielding), the subrequest fails to wake up its parent request, causing the parent's Lua coroutine to hang indefinitely.

Root cause: HTTP/2 sets c->error=1 immediately on client disconnect, triggering the error path in ngx_http_finalize_request(). This path calls ngx_http_terminate_request(), which clears the posted_requests queue before the parent can be processed.

Solution: In ngx_http_finalize_request(), when a subrequest completes with c->error set, manually update subrequest state and wake the parent request, then return immediately to avoid ngx_http_terminate_request().

This fix ensures the parent request can process the subrequest response even when the client connection is closed, which is critical for OpenResty's async subrequest mechanism with ngx.location.capture().

fix: #1076

I hereby granted the copyright of the changes in this pull request
to the authors of this openresty project.

PR #1132, opened 2026-06-13
bugfix: limit HTTP/2 subrequest error wakeup

oowl:

## Summary

This narrows the `http2_subreq_error_wakeup` nginx patches so HTTP/2 subrequests only skip `ngx_http_terminate_request()` when they have a `post_subrequest` callback.

PR #1083 needs that fallthrough for callback-driven subrequests such as `ngx.location.capture`, where the parent request must be woken instead of having its posted requests cleared by termination. Native nginx subrequests without a callback, such as slice range subrequests, do not need that wakeup path.

## Root Cause

The current guard only checks for HTTP/2 subrequests:

```c
if (r->http_version < NGX_HTTP_VERSION_20 || r == r->main) {
    ngx_http_terminate_request(r, rc);
    return;
}
```

That also catches native subrequests with `r->post_subrequest == NULL`. With HTTP/2 + `slice` + `proxy_cache`, a stalled or aborting client can finalize a slice subrequest with `NGX_HTTP_REQUEST_TIME_OUT`. Instead of terminating, it falls through to `ngx_http_special_response_handler()`, tries to send a 408 header after the 200 header was already sent, and floods the error log with `header already sent`.

## Changes

All existing `nginx-*-http2_subreq_error_wakeup.patch` files now terminate HTTP/2 subrequest errors when the subrequest has no callback:

```c
if (r->http_version < NGX_HTTP_VERSION_20
    || r == r->main
    || r->post_subrequest == NULL)
{
    ngx_http_terminate_request(r, rc);
    return;
}
```

This keeps the callback wakeup behavior from PR #1083 while excluding native subrequests such as slice.

Test case: https://github.com/openresty/lua-nginx-module/pull/2507

### DIFF
--- a/patches/nginx/1.27.1/nginx-1.27.1-http2_subreq_error_wakeup.patch
+++ b/patches/nginx/1.27.1/nginx-1.27.1-http2_subreq_error_wakeup.patch
@@ -2,12 +2,15 @@ diff --git a/src/http/ngx_http_request.c b/src/http/ngx_http_request.c
 index 9593b7f..dbc23ff 100644
 --- a/src/http/ngx_http_request.c
 +++ b/src/http/ngx_http_request.c
-@@ -2560,8 +2560,15 @@ ngx_http_finalize_request(ngx_http_request_t *r, ngx_int_t rc)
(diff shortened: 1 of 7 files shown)


oowl:
close dut to duplicated one https://github.com/openresty/openresty/pull/1133

